PR #24, opened 2020-05-27
Skip deleted packages in git changes

### LINKED ISSUE #40: Can't remove v0 of a package
#40061 is a good example that tries to remove v0 of puppeteer. I think this is something that explicitly needs to be allowed, where we previously thought it was fine to keep the old versions around.

sandersn:
The actual change is just `if (diff.status === "D") continue` plus a test; everything else is prettier diffs.

Fixes #40


jakebailey:
Old old PR, but I think this all is no longer needed given recent changes?

sandersn:
Yes, in fact we ended up going the opposite direction: only using git to track deletions.

PR #30, opened 2020-06-04
Fix crash in inverted version packages

sandersn:
Packages with inverted versions -- where the toplevel index.d.ts has a lower version than some /v[0-9]+ subdirectory -- are unexpected by generate-packages. It expects that non-latest versions are always in a subdirectory.

This is not true for react-bootstrap, and no other packages that I'm aware of. I can't remember how react-bootstrap got into this state, but it should not stay there until the issue is actually fixed.

This is just a crash fix, not a complete fix. There aren't any tests because I didn't want to have to add testing infrastructure for generateTypingPackage on short notice -- all the existing tests just test that the correct file contents are generated, not the structure of the generated package itself.

This is not really a complete change, since I'm trying to get a fix out quickly. Either CI needs to prevent this state from happening, or publishing needs to support it.

### DIFF
--- a/packages/publisher/.vscode/launch.json
+++ b/packages/publisher/.vscode/launch.json
@@ -9,7 +9,7 @@
             "request": "launch",
             "name": "calculate-versions.js",
             "cwd": "${workspaceFolder}",
-            "program": "${workspaceFolder}/bin/calculate-versions.js",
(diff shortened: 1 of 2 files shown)


sandersn:
oh good, `guardian.exe` is missing for some reason now.

PR #32, opened 2020-06-04
Use standard guardian tasks, not a manual script

sandersn:
The script used a specific version, but guardian revs all the time.


sandersn:
Well, never mind. The built-in tasks pass a bogus (outdated?) flag `--severity` so I just used `dir` with a glob to run all the versions on disk, which will hopefully be only one.

PR #51, opened 2020-06-04
Delete custom logging to blob storage and webhook

andrewbranch:
- Nobody looks at these logs.
- The webhook occasionally allows packages to be published a maximum of 33 minutes sooner than it would have without the webhook.


packages/publisher/README.md
```diff
@@ -317,62 +316,6 @@ Scripts should save this log under a unique filename so any errors may be review
...
-
-* Set up forwarding:
-	* Install [ngrok](https://ngrok.com)
-	* `ngrok http 80` (or whatever `PORT` environment variable you're using)
-	* Copy the forwarding URL (Looks like: http://deadbeef.ngrok.io)
-
-* Add a hook:
-	* Go to https://github.com/your/dummy-repo/settings/hooks
-	* Payload URL = url copied from ngrok
-	* Secret = make up a word here
```
sandersn, line 348: that's another way to get rid of "swordfish"...

packages/publisher/README.md
```diff
@@ -408,32 +351,12 @@ Optional. Example value `latest`
 
 If present, packages are published with the provided version tag.
 
-### azureStorageAccount
```
sandersn, line 411: so...we can drop our azure storage account too?

packages/publisher/src/main.ts
```diff
@@ -0,0 +1,93 @@
...
+        properties: {
+          error: e.toString()
+        }
+      });
+      throw e;
+    }
+  }
+}
+
+// Even if there are many changes to DefinitelyTyped in a row, we only perform one update at a time.
```
sandersn, line 68: this comment should instead talk about very large updates

packages/publisher/src/main.ts
```diff
@@ -0,0 +1,93 @@
...
+    }
+    const fetcher = new Fetcher();
+
+    try {
+      const fullOne = updateOneAtATime(async (log, timestamp) => {
+        log.info("");
+        log.info("");
+        log.info(`# ${timestamp}`);
+        log.info("");
+        log.info("Starting full...");
```
sandersn, line 28: one thing to test is whether the app service actually properly starts without a web request coming in. Not actually running a web server may confuse the App Service machinery.

packages/publisher/src/main.ts
```diff
@@ -0,0 +1,93 @@
...
+          const result = fullOne(log, timestamp);
+          if (!result) {
+            return;
+          } // already working, so do nothing.
+          result.catch(e => {
+            log.info(e.toString());
+            console.error(e);
+          });
+        },
+        2_000_000,
```
sandersn, line 52: need to make sure that an initial runs happens immediately instead of waiting 2000 s *first*. I've observed some bad behaviour in the past, but it may be a result of my previous comment.

packages/publisher/README.md
andrewbranch: I believe it’s no longer used, but could be useful in the future (we can automatically send application logs there, for example)

packages/publisher/src/main.ts
andrewbranch: You’re right, I think this was wrong. Fixed.

karlhorky:
Oh it looks like this search index is being relied upon by Algolia's `npm-search` index:

https://github.com/algolia/npm-search/blob/d99c31fbe733953d39e498f9c7d982b2e09f3458/src/typescript/index.ts#L24-L40

Is there a recommendation for getting this types information from somewhere else? Eg.

1. Is a package written in TypeScript?
2. If not, does the package have a corresponding `@types/*` package published to Definitely Typed?

Seems like there was also [TypeSearch](https://www.typescriptlang.org/dt/search) (and the [TypeScout](https://typescout.vercel.app/) interface for it), but it seems like this is either unavailable or not containing all information that is necessary.

cc @Haroenv

PR #52, opened 2020-06-04
Update @azure/cosmos to 3

sandersn:
This gets rid of a transitive dependency on md5.js

The biggest change is that `query` returns a nested iterator, so it exposes the chunking nature of requesting results.

I think. That's inferred from the type changes. This passes all the tests, but I'm not sure if that's sufficient.

### DIFF
--- a/packages/perf/src/cli/compareTypeScript.ts
+++ b/packages/perf/src/cli/compareTypeScript.ts
@@ -232,17 +232,15 @@ async function getPackagesToTestAndPriorResults(
   packageList?: PackageId[]
 ) {
   const iterator: AsyncIterable<FeedResponse<QueryResult<Document<PackageBenchmarkSummary>>>> = await container.items
-    .query(
-      {
-        query:
-          `SELECT * FROM ${config.database.packageBenchmarksContainerId} b` +
-          `  WHERE b.body.typeScriptVersionMajorMinor = @typeScriptVersion` +
-          (packageList
-            ? `  AND b.body.packageName IN (${packageList!.map(({ name }) => `"${name}"`).join(", ")})` // Couldn’t figure out how to do this with parameters
-            : ""),
-        parameters: [{ name: "@typeScriptVersion", value: typeScriptVersion }]
-      }
-    )
+    .query({
+      query:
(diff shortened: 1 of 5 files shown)


packages/perf/src/cli/compareTypeScript.ts
```diff
@@ -231,55 +231,53 @@ async function getPackagesToTestAndPriorResults(
...
-        query:
-          `SELECT * FROM ${config.database.packageBenchmarksContainerId} b` +
-          `  WHERE b.body.typeScriptVersionMajorMinor = @typeScriptVersion` +
-          (packageList
-            ? `  AND b.body.packageName IN (${packageList!.map(({ name }) => `"${name}"`).join(", ")})` // Couldn’t figure out how to do this with parameters
-            : ""),
-        parameters: [{ name: "@typeScriptVersion", value: typeScriptVersion }]
-      },
-      {
-        enableCrossPartitionQuery: true
```
andrewbranch, line 246: I went hunting for this, and it looks like cross-partition queries are the default now, but I wonder if @southpolesteve could confirm—he was on most of the PRs around the v3 SDK update.

sandersn: I found text To that effect yesterday. No link, sorry — I’m on my phone right now.

southpolesteve: This is correct. No flag anymore. Queries will go cross partition automatically if needed.

andrewbranch: Thanks @southpolesteve! This is a nice API change. Our partitions are small so we almost always had to pass this flag.

southpolesteve: Glad you like it! The feedback has been the opposite for people with a lot of large partitions. They like explicit as it can be a perf foot gun. No plans yet for a Cosmos v4, but if we do one, I might look at splitting out cross partition query into its own top-level package function. Something like:
```js

import { crossPartitionQuery, CosmosClient } 

const client = new CosmosClient()
const container = client.database('id').container('id')

const queryIterator = crossPartitionQuery(container, "SELECT * from c")
```

PR #54, opened 2020-06-05
Move @type dependencies to devDependencies

davidaurelio:
Moves the following dependencies to `devDependencies`

- in `@dt/utils`: `@types/node`
- in `@dt/header-parser`: `@types/parsimmon`

## Motivation

When using `dtslint` in a project, these two packages get pulled in, and consequently also the `@types/…` packages that they depend on.

Additional types in `node_modules` affect type checking, with unintended consequences:

- for projects that have their own definition of the global `require`, that is going to clash with the definition in `@types/node`
- if typescript 2.x support is desired, `@types/parsimmon` will trigger an error (TS2370: A rest parameter must be of an array type).

Looking at the source, it looks as if these types are used to support development of `@dt/utils`/ `@dt/header-parser` themselves, so moving them seemed like a viable solution.

### DIFF
--- a/packages/header-parser/package.json
+++ b/packages/header-parser/package.json
@@ -21,8 +21,10 @@
   },
   "dependencies": {
     "@definitelytyped/typescript-versions": "^0.0.38",
-    "@types/parsimmon": "^1.10.1",
     "parsimmon": "^1.13.0"
+  },
+  "devDependencies": {
+    "@types/parsimmon": "^1.10.1"
   },
   "publishConfig": {
     "access": "public"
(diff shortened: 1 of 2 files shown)


davidaurelio:
Well, I will have to run the cla by my employer first. That’s gonna take a bit. Yay!

andrewbranch:
Sorry! 😭

davidaurelio:
Haha, I have been on both sides of this :)

jakebailey:
It's been a few years; I'm not sure if this is still an issue but parsimmon is gone now, and overall there are conflicts. It seems to me that we have to have the dep on node given we import it and export it... I can see a world in which we drop the need for the stream stuff, but we also export `exec` helpers which depend on the big exec options types.

I think this PR itself is unlikely to be revived, though, but I'm still curious if this is a problem. It seems strange to depend on our package's APIs but also declare a custom require; we definitely only expect to be run within node...

PR #57, opened 2020-06-09
Deduplicate dependencies

sandersn:
Use the more common object format, which is technically less powerful, but we've never needed the additional power of the array format.


sandersn:
Ported from microsoft/types-publisher#771. I had to propagate the changes out to the perf package, which would have had to happen before, but now it's immediate since they're in the same monorepo.

PR #58, opened 2020-06-09
Rename dependenciesWhitelist to allowablePackageJsonDependencies

andrewbranch:
I’ve heard from some people that the term “whitelist” is problematic. Additionally, I can never remember what the dependencies in this list are allowed to do, so renaming it to something more descriptive feels like a no-brainer. I wanted to go ahead and do it now because we link to this file from the old types-publisher repo README, and we’re getting to make that repo read-only, so these steps need to happen in order.


andrewbranch:
> the adjective "allowed" is way more frequent and just as applicable as "allowable"

Fair enough. I’ll allowable it.

PR #63, opened 2020-06-23
docs: updated repository links

eps1lon:
Otherwise it's not obvious where the source lives for the corresponding npm packages.

Aside: Why is it owned by microsoft and not in the DT org? Will this be moved later and the links already anticipated this move or are the links outdated in which case: why did it move?

### DIFF
--- a/packages/definitions-parser/package.json
+++ b/packages/definitions-parser/package.json
@@ -2,13 +2,13 @@
   "name": "@definitelytyped/definitions-parser",
   "version": "0.0.40",
   "description": "Reads the DefinitelyTyped repository and provides an API for querying its metadata",
-  "homepage": "https://github.com/DefinitelyTyped/tools#readme",
+  "homepage": "https://github.com/microsoft/DefinitelyTyped-tools/tree/master/packages/definitions-parser#readme",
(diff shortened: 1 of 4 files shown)


sandersn:
@andrewbranch may be able to explain it better than I can, but briefly:

- The packages in this monorepo are for Microsoft's publishing of `@types` packages from Definitely Typed. Microsoft publishes these packages and is responsible for them. That's *technically* different than Definitely Typed, which is not legally connected to Microsoft. 
- Historically, most of the code here was previously in microsoft/types-publisher.
- Other parts of the repo track the currently supported versions of Typescript and need to update as part of the Typescript release process.

This repo is (probably) not intrinsically tied to the microsoft org, but there are a lot of small reasons why it's here.

andrewbranch:
To my knowledge, the only reason this is in the Microsoft org is that it contains code that deploys directly to a Microsoft internal Azure subscription, and so the contributor access needs to be managed solely by Microsoft. Personally I would love for the original DT maintainers to be able to have more autonomy within this repo, but ultimately we figured having an employee be the one to click merge buttons shouldn’t be a big problem for anyone else who wants to contribute.

PR #65, opened 2020-06-29
Add timeout to downloadAndExtractFile

### LINKED ISSUE #64: Failing to clone Definitely Typed wedges the publisher
Sometimes it appears that Definitely Typed fails to clone, but the clone never times out, and the publisher is stuck waiting forever.

Instead, the clone should wait no more than one or two thousand seconds before timing out and killing the process (or waiting until next time).

sandersn:
It's only used to download the tgz of DT's master.

I set the timeout to 1000 seconds and added an error-handling line? I don't actually know how node's error handling works so I need to go and learn the right way to do it.

Fixes #64

### DIFF
--- a/packages/utils/src/io.ts
+++ b/packages/utils/src/io.ts
@@ -184,9 +184,10 @@ export function downloadAndExtractFile(url: string): Promise<FS> {
       dir.set(baseName, content);
     }
 
-    https.get(url, response => {
+    https.get(url, { timeout: 1_000_000 }, response => {
       const extract = tarStream.extract();
       response.pipe(zlib.createGunzip()).pipe(extract);
+      response.on("error", reject);
       interface Header {
         readonly name: string;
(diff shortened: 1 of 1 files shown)


packages/utils/src/io.ts
```diff
@@ -184,9 +184,10 @@ export function downloadAndExtractFile(url: string): Promise<FS> {
       dir.set(baseName, content);
     }
 
-    https.get(url, response => {
+    https.get(url, { timeout: 1_000_000 }, response => {
       const extract = tarStream.extract();
       response.pipe(zlib.createGunzip()).pipe(extract);
+      response.on("error", reject);
```
sandersn, line 190: further reading suggests that the right place to put this is on the return value of `https.get`, not on `response`. Anybody care to confirm that?

sandersn:
Argh, prettier! Ignore whitespace changes in this review.

PR #79, opened 2020-08-07
Allow moment-timezone as package.json dependency

chivesrs:
moment-timezone now serves its own typings. Add it here so the DT typings can be deleted.

See https://github.com/DefinitelyTyped/DefinitelyTyped/pull/46421


sandersn:
It's now published, so Definitely Typed's CI will get the new list next time it runs.

PR #80, opened 2020-08-08
feat: allow bee-queue for bull-arena types

skeggse:
For https://github.com/DefinitelyTyped/DefinitelyTyped/pull/46626, which references types from `bee-queue`.

(Disclaimer: I'm the primary maintainer for both packages)


sandersn:
It's now published, so Definitely Typed's CI will get the new list next time it runs.

PR #85, opened 2020-08-19
Add support for 4.1, ship 4.0, remove 3.0 support

sandersn:
I forgot to push this PR last week when I retagged the existing `@types` packages for the RC.

This means that I'll have to retag the packages again, like I did last year before I fixed the process.

I'll actually publish this when 4.0 ships tomorrow.


sandersn:
Weird, local `yarn test` missed the build error in header-parser.

PR #91, opened 2020-08-26
Flip typesVersions order to oldest-first

sandersn:
typesVersions subfolders are now expected to cover *older* versions with the root having the newest version, and the entry in package.json should be ordered oldest-first now:

For example, a package with 3.2,3.5,3.6 covers these ranges:
3.1-3.2, 3.3-3.5, 3.6, 3.7-4.1

And has the following typesVersions keys:

<=3.2, <=3.5, <=3.6

Publishing this change will block changes to the 18 DT packages that still have typesVersions entries, so I will fix them locally using the prerelease package once this PR is merged.

Edit: Before this change, a package with 3.2,3.5,3.6 covers different ranges and has different typesVersions keys:

3.1, 3.2-3.4, 3.5, 3.6-4.1

with keys:

>=3.6, >=3.5, >=3.2

### DIFF
--- a/packages/header-parser/src/index.ts
+++ b/packages/header-parser/src/index.ts
@@ -41,11 +41,11 @@ export function makeTypesVersionsForPackageJson(typesVersions: readonly TypeScri
     return undefined;
   }
 
-  const newestFirst = typesVersions.slice();
-  newestFirst.sort((v1, v2) => (v1 < v2 ? 1 : v1 > v2 ? -1 : 0));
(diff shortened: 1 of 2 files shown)


sandersn:
After this change is published, dtslint will need to update to the new version.

sandersn:
Actually, if prerelease packages deploy to types-publisher, then this shouldn't be merged until the 18 packages have PRs ready to go.

PR #94, opened 2020-08-27
Deduplicate files list

jablko:


packages/definitions-parser/src/lib/definition-parser.ts
```diff
@@ -200,9 +200,15 @@ async function combineDataForAllTypesVersions(
...
 
-  const files = Array.from(
+  const files = unique(
     flatMap(allTypesVersions, ({ typescriptVersion, declFiles }) =>
-      declFiles.map(file => (typescriptVersion === undefined ? file : `ts${typescriptVersion}/${file}`))
+      declFiles.map(file =>
+        typescriptVersion === undefined
+          ? file
+          : file.startsWith("../")
+          ? file.slice(3)
```
sandersn, line 209: this is new, right? Can you add a test to show this off too?

sandersn (changes requested):
Seems like a good idea, but can you add an explanation of when it would be useful, or maybe a test based on a real occurrence of dupes?

jakebailey:
This will no longer be needed after #769, which replaces this code with plain globbing.

PR #99, opened 2020-09-01
Keep local references out of testDependencies

jablko:
microsoft/types-publisher#777


jakebailey:
`testDependencies` is no more after the pnpm migration.

PR #100, opened 2020-09-01
Handle relative type references

jablko:
microsoft/types-publisher#776

### DIFF
--- a/packages/definitions-parser/src/lib/module-info.ts
+++ b/packages/definitions-parser/src/lib/module-info.ts
@@ -225,8 +225,7 @@ function findReferencedFiles(src: ts.SourceFile, packageName: string, subDirecto
     addReference({ text: ref.fileName, exact: true });
   }
   for (const ref of src.typeReferenceDirectives) {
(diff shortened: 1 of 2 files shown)


jablko:
> If I recall correctly, this rule is intentional; of the examples you linked to, only the nested reference in slickgrid should be allowed, so probably the change should be to require as many leading ../ as the number of nested directories + 1.

From [the history](https://github.com/microsoft/types-publisher/commit/c1695235568b5c286f888b2c78e1a6fe7f2bf609) it looks like when types-publisher only had support for non-relative type references, only `/// <reference types="packagename/x" />` were local, which makes sense, and this was applied to relative type references by analogy, but that doesn't make sense? `addReference()` already correctly handles non-local relative references: https://github.com/microsoft/DefinitelyTyped-tools/blob/5e8a23e6b9a90797f64b67f6b07eb8e069d979a1/packages/definitions-parser/src/lib/module-info.ts#L253-L262

sandersn:
Fixup PR is https://github.com/DefinitelyTyped/DefinitelyTyped/pull/44382

jakebailey:
This code's going away in #788.

PR #103, opened 2020-09-02
Self references are local

jablko:


jakebailey:
`module-info` is going away entirely in #788, closing.

PR #104, opened 2020-09-02
Remove internal paths from test dependencies

jablko:
Fixes microsoft/types-publisher#517


jakebailey:
Now that DT is a pnpm monorepo, `getTestDependencies` is no more. Closing.

PR #105, opened 2020-09-02
Keep self out of testDependencies

jablko:


jakebailey:
Now that DT is a pnpm monorepo, `getTestDependencies` is no more. Closing.

PR #108, opened 2020-09-03
Exclude test dependents' dependents from affected

jablko:
If
- B depends on A
- C depends on B

and
- X test-depends on A
- D depends on X

and A changes, then B, C and X are also affected, but not D. D doesn't reference A anywhere in its dependencies.


jakebailey:
This is all now largely replaced by explicit the dependency graph provided by pnpm / package.json.

PR #109, opened 2020-09-03
Fix '--clone' vs '--path' conflict due to defaults

rbuckton:
If `--path` has a default, then it will always conflict with `--clone`, which is causing problems with the On-Demand DT tests for TypeScript


andrewbranch (approved):
👍

PR #112, opened 2020-09-04
Exclude dependencies' test dependencies from transitive dependencies

jablko:
If
- A test-depends on X

and
- X depends on B
- X test-depends on Y

then A's tests reference X and B but not Y.


jakebailey:
This stuff has been replaced by `pnpm ls`; closing.

PR #116, opened 2020-09-14
Deprecate Typescript 3.1

sandersn:
[According to the schedule](https://github.com/DefinitelyTyped/DefinitelyTyped#older-versions-of-typescript-30-and-earlier), it's time to deprecate 3.1. It seems like not much time since I delayed the 3.0 deprecation to match the 4.0 release.

Removing 3.1 support was complicated. This should be much easier. The follow-up tasks are:

1. Update dtslint and dts-critic.
2. Remove ts3.1/ directories from web-animations-js, openpgp, jest, es-to-primitive.
3. Remove ts3./1 directories from all versions of node. (This is big enough that it deserves its own task; the default implementations will have to move up to the next TS version directory since node's types build up from the oldest versions.)
4. Update "lowest version number" in DT's README.

### DIFF
--- a/packages/header-parser/src/index.ts
+++ b/packages/header-parser/src/index.ts
@@ -230,7 +230,7 @@ const typeScriptVersionLineParser: pm.Parser<AllTypeScriptVersion> = pm
 const typeScriptVersionParser: pm.Parser<AllTypeScriptVersion> = pm
   .regexp(/\r?\n/)
   .then(typeScriptVersionLineParser)
-  .fallback<TypeScriptVersion>("3.2");
+  .fallback<TypeScriptVersion>(TypeScriptVersion.shipped[0]);
 
 export function parseTypeScriptVersionLine(line: string): AllTypeScriptVersion {
   const result = typeScriptVersionLineParser.parse(line);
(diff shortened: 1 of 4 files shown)


sandersn:
Note that follow-up task (3) should happen after DefinitelyTyped/DefinitelyTyped#47576 is merged.

packages/header-parser/src/index.ts
```diff
@@ -230,7 +230,7 @@ const typeScriptVersionLineParser: pm.Parser<AllTypeScriptVersion> = pm
 const typeScriptVersionParser: pm.Parser<AllTypeScriptVersion> = pm
   .regexp(/\r?\n/)
   .then(typeScriptVersionLineParser)
-  .fallback<TypeScriptVersion>("3.1");
+  .fallback<TypeScriptVersion>("3.2");
```
andrewbranch, line 233: Any reason not to do `TypeScriptVersion.shipped[0]`?

andrewbranch (approved):

packages/header-parser/src/index.ts
sandersn: Good idea!

PR #120, opened 2020-09-21
Generalize #92 to deep imports from older versions

jablko:
Checks that wildcard path mappings are present when necessary. E.g. if you have a `"react-dom": ["react-dom/v15"]` path mapping and import `react-dom/server`, this checks that a wildcard `react-dom/*` path mapping is also present.


jakebailey:
Dependencies are now explicit post-pnpm, so this is no longer needed.

PR #121, opened 2020-09-22
Try to get minTypeScriptVersion-compatible dependencies

jablko:
We'll be testing the types with `minTypeScriptVersion` (modulo `onlyTestTsNext`) so make an effort to get compatible external dependencies by adding the `--tag ts${minTypeScriptVersion}` option to `npm install`. If a dependency provides such a tag then we'll get that compatible version. If not, well, we'll continue to get the current behavior.

This adds the `--tag` option to `prepareAffectedPackages()` (trivial) and `prepareAllPackages()`, which previously used separate logic to crawl the types directory and bypassed `@definitelytyped/definitions-parser`. Adding the `--tag` option means reading the `minTypeScriptVersion`, so I made `prepareAffectedPackages()` and `prepareAllPackages()` share the same logic. I think this is desirable: You should get the same per-package behavior regardless of `onlyRunAffectedPackages`?

### DIFF
--- a/packages/dtslint-runner/src/prepareAllPackages.ts
+++ b/packages/dtslint-runner/src/prepareAllPackages.ts
@@ -1,52 +1,31 @@
-import fs from "fs";
-import { pathExists } from "fs-extra";
-import { joinPaths, nAtATime, execAndThrowErrors, npmInstallFlags } from "@definitelytyped/utils";
+import {
+  AllPackages,
+  getDefinitelyTyped,
+  checkParseResults,
+  parseDefinitions
+} from "@definitelytyped/definitions-parser";
+import { joinPaths, loggerWithErrors } from "@definitelytyped/utils";
+import { installDependencies } from "./prepareAffectedPackages";
 import { PreparePackagesOptions, PreparePackagesResult } from "./types";
 
 export async function prepareAllPackages({
   definitelyTypedPath,
-  noInstall
+  noInstall,
+  nProcesses
 }: PreparePackagesOptions): Promise<PreparePackagesResult> {
   const typesPath = joinPaths(definitelyTypedPath, "types");
-  const allPackages = await getAllPackages(typesPath);
+  const [log] = loggerWithErrors();
+  const options = {
+    definitelyTypedPath,
+    progress: false,
+    parseInParallel: nProcesses > 1
+  };
+  const dt = await getDefinitelyTyped(options, log);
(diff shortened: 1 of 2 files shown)


packages/dtslint-runner/src/prepareAllPackages.ts
```diff
@@ -1,52 +1,31 @@
...
   const typesPath = joinPaths(definitelyTypedPath, "types");
-  const allPackages = await getAllPackages(typesPath);
+  const [log] = loggerWithErrors();
+  const options = {
+    definitelyTypedPath,
+    progress: false,
+    parseInParallel: nProcesses > 1
+  };
+  const dt = await getDefinitelyTyped(options, log);
+  await parseDefinitions(dt, nProcesses ? { definitelyTypedPath, nProcesses } : undefined, log);
```
andrewbranch, line 24: This is expensive and unnecessary when testing all packages, which is why it wasn’t here before.

andrewbranch (changes requested):

packages/dtslint-runner/src/prepareAllPackages.ts
jablko: :+1: That line (`parseDefinitions()`) is easy enough to ax, however it looks like it's needed in order to get the `minTypeScriptVersion` of each package?

andrewbranch: Ah, I see. If that’s the case then it’s probably worth it. From the description I thought you were consolidating just for code reuse, so I wanted to be clear that there’s a pretty big performance tradeoff.

jablko: I could get the `minTypeScriptVersion` a lot more efficiently than by calling `parseDefinitions()`, but on the other hand testing all packages is already an expensive operation. Like off the top of my head it currently takes about a couple hours, of which +/- 60 seconds is spent in `parseDefinitions()`? How does this optimization pay off?

andrewbranch (approved):

packages/dtslint-runner/src/prepareAllPackages.ts
andrewbranch: Yeah, you’re right. My bad 👍

andrewbranch:
I want to make sure this makes sense to @sandersn first

jablko:
Absolutely, thanks for your help :+1:

sandersn (approved):

PR #122, opened 2020-09-22
Use libraryName for deprecation dependency

sandersn:
Previously, the code incorrectly re-used the `@types` package name for the deprecation dependency, which is also noted in the readme as the correct package to install in the future.

This is correct 90%+ percent of the time because `@types/X` usually provides types for package `X`. Sometimes, however, there is no matching `X`, so that `@types/X` should tell users to install some other package `Y` when it's deprecated.

For example, `@types/xumm-api` was for a non-npm package that eventually got published to npm as `xumm-sdk` and at the same time started shipping types. So the deprecation entry looked like this:

```json
{
  "libraryName": "xumm-sdk",
  "typingsPackageName": "xumm-api",
  "sourceRepoURL": "https://github.com/XRPL-Labs/XUMM-SDK",
  "asOfVersion": "0.1.4"
}
```

There IS no `xumm-api`, only an `xumm-sdk`, so it doesn't make sense to point people to `xumm-api` when `@types/xumm-api` is deprecated.

### DIFF
--- a/packages/definitions-parser/src/packages.ts
+++ b/packages/definitions-parser/src/packages.ts
@@ -288,12 +288,12 @@ export class NotNeededPackage extends PackageBase {
   }
 
   readme(): string {
-    return `This is a stub types definition for ${this.libraryName} (${this.sourceRepoURL}).\n
+    return `This is a stub types definition for ${getFullNpmName(this.name)} (${this.sourceRepoURL}).\n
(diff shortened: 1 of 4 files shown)


andrewbranch (approved):

sandersn:
I'm going to ship a new version and see whether I can properly deprecate `@types/xumm-api`.

PR #124, opened 2020-09-23
Cosmetic

jablko:
Makes this path juggling easier to read, I think?

### DIFF
--- a/packages/definitions-parser/src/lib/module-info.ts
+++ b/packages/definitions-parser/src/lib/module-info.ts
@@ -253,10 +253,7 @@ function findReferencedFiles(src: ts.SourceFile, packageName: string, subDirecto
       full = full.slice(packageName.length + 4);
     } else if (baseDirectory && full.startsWith("../" + baseDirectory + "/")) {
       full = full.slice(baseDirectory.length + 4);
-    } else if (
-      full.startsWith("..") &&
-      (baseDirectory === "" || path.normalize(joinPaths(baseDirectory, full)).startsWith(".."))
-    ) {
+    } else if (full.startsWith(baseDirectory ? "../.." : "..")) {
       throw new Error(
         `${src.fileName}: ` +
           'Definitions must use global references to other packages, not parent ("../xxx") references.' +


packages/definitions-parser/src/lib/module-info.ts
```diff
@@ -253,10 +253,7 @@ function findReferencedFiles(src: ts.SourceFile, packageName: string, subDirecto
       full = full.slice(packageName.length + 4);
     } else if (baseDirectory && full.startsWith("../" + baseDirectory + "/")) {
       full = full.slice(baseDirectory.length + 4);
-    } else if (
-      full.startsWith("..") &&
-      (baseDirectory === "" || path.normalize(joinPaths(baseDirectory, full)).startsWith(".."))
-    ) {
+    } else if (full.startsWith(baseDirectory ? "../.." : "..")) {
```
sandersn, line 256: Is this equivalent? When `baseDirectory !== ""`, the original checks `joinPaths(baseDirectory, full)`, but the new version checks `full` with a longer initial string.

This could be obviously correct if `full` were produced from `baseDirectory`, but it's produced from `subDirectory`. Is there some equivalence of those two that I'm missing?

sandersn (changes requested):

packages/definitions-parser/src/lib/module-info.ts
jablko: Assuming `full` is already normalized (it is), then the following are equivalent
- `path.normalize(joinPaths("segment", full)).startsWith("..")`
- `full.startsWith("../..")`

If ``path.normalize(`segment/${full}`)`` starts with `".."` then `full` must have started with `"../.."`, to have soaked up `"segment"` and still start with `".."`, and if `full` starts with `"../.."` then ``path.normalize(`segment/${full}`)`` will still start with `".."`.

So, as long as `baseDirectory` isn't `"."` or `".."` (it isn't, it's `"ts2.0"` or the like), then this change is equivalent.

sandersn:
Thanks for the explanation. Unfortunately, I thought about it more and I don't want to merge a cosmetic change to a complex piece of code. I'm too worried about failures for little benefit. (A lot of it is that I swapped the details of this code out of my head a long time back.)

sandersn:
@andrewbranch points out that a really valuable addition to this code would be detailed documentation, whether it's inline or in a separate document with a pointer in the source.

dieguesmosken (commented):
its fine

jakebailey:
This code's going away in #788, replaced with a lint rule.

PR #127, opened 2020-10-02
Move isRelativePath() from dt-mergebot

jablko:
Put it together with the other `checkFilesFromTsConfig()`/`OTHER_FILES.txt` checks.

### DIFF
--- a/packages/definitions-parser/src/lib/definition-parser.ts
+++ b/packages/definitions-parser/src/lib/definition-parser.ts
@@ -278,6 +278,11 @@ function getTypingDataForSingleTypesVersion(
           .split(/\r?\n/g)
           .filter(Boolean)
       : [];
+  for (const file of otherFiles) {
+    if (!isRelativePath(file)) {
+      throw new Error(`In ${packageName}: A path segment is empty or all dots ${file}`);
+    }
+  }
   checkAllFilesUsed(ls, usedFiles, otherFiles, packageName, fs);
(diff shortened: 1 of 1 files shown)


packages/definitions-parser/src/lib/definition-parser.ts
```diff
@@ -399,6 +407,10 @@ Other d.ts files must either be referenced through index.d.ts, tests, or added t
   }
 }
 
+function isRelativePath(path: string) {
+  return path.split(/\//).every(part => part.length > 0 && !part.match(/^\.+$|[\\\n\r]/));
```
andrewbranch, line 411: I get what the point is, but this returns true for absolute paths, it looks like

jablko: I extracted it verbatim from https://github.com/DefinitelyTyped/dt-mergebot/blob/b8d318e166bc44484242cdf0bad7fea8c020fa45/src/pr-info.ts#L418-L420 and I think it works for absolute paths, because `"/absolute".split(/\//)` -> `["", "absolute"]`, so `"/absolute".split(/\//).every(part => part.length > 0)` -> `false`. Unless I'm missing something?

andrewbranch: Ah, I forgot the first element would be an empty string. Works for me.

jablko: Thanks!

PR #128, opened 2020-10-04
Add `html-webpack-plugin`

peterblazejewicz:
There are couple of definitions that depends on `html-webpack-plugin`
and using old definition type, instead of using inline version shipped
with v4.0.0 of plugin itself.

https://github.com/jantimon/html-webpack-plugin/blob/master/typings.d.ts

Thanks!

### DIFF
--- a/packages/definitions-parser/allowedPackageJsonDependencies.txt
+++ b/packages/definitions-parser/allowedPackageJsonDependencies.txt
@@ -120,6 +120,7 @@ handlebars
 helmet
 hoist-non-react-statics
 hot-shots
+html-webpack-plugin
 htmlparser2
 i18next
 immer


sandersn:
I take it those definitions need a package.json in order to depend on html-webpack-plugin? Which ones are they?

peterblazejewicz:
Indirect is `react-scripts`:
https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/package.json#L58
and then, in the DT world, these are:
`react-dev-utils`, introducing depeendency to `html-webpack-plugin` via type imports:
https://github.com/facebook/create-react-app/blob/master/packages/react-dev-utils/InlineChunkHtmlPlugin.js
https://github.com/facebook/create-react-app/blob/master/packages/react-dev-utils/InterpolateHtmlPlugin.js
another webpack plugin `csp-html-webpack-plugin`, this time as proper dependency:
https://github.com/slackhq/csp-html-webpack-plugin/blob/master/package.json#L38
`html-webpack-template` - type imports:
https://www.typescriptlang.org/dt/search?search=html-webpack-template
`poi` as type import:
https://www.typescriptlang.org/dt/search?search=poi

I'm not sure how to apprach a problem of upgrading. For now I'm only interested in getting `html-webpack-plugin` upgraded itself for our own webpack deps, but this could wait.

sandersn (approved):

PR #136, opened 2020-10-16
Move empty OTHER_FILES.txt check from dt-mergebot

jablko:
Like #127, put [this check](https://github.com/DefinitelyTyped/dt-mergebot/blob/72404d3fb1af96cff27d8e781c03a84092ead1e9/src/pr-info.ts#L397) together with the others.

### DIFF
--- a/packages/definitions-parser/src/lib/definition-parser.ts
+++ b/packages/definitions-parser/src/lib/definition-parser.ts
@@ -270,14 +270,16 @@ function getTypingDataForSingleTypesVersion(
     packageDirectory
   );
   const usedFiles = new Set([...types.keys(), ...tests.keys(), "tsconfig.json", "tslint.json"]);
-  const otherFiles =
-    ls.indexOf(unusedFilesName) > -1
-      ? fs
-          // tslint:disable-next-line:non-literal-fs-path -- Not a reference to the fs package
-          .readFile(unusedFilesName)
-          .split(/\r?\n/g)
-          .filter(Boolean)
-      : [];
+  const otherFiles = ls.includes(unusedFilesName)
+    ? fs
+        // tslint:disable-next-line:non-literal-fs-path -- Not a reference to the fs package
+        .readFile(unusedFilesName)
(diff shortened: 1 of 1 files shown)


packages/definitions-parser/src/lib/definition-parser.ts
```diff
@@ -270,14 +270,16 @@ function getTypingDataForSingleTypesVersion(
...
-          .filter(Boolean)
-      : [];
+  const otherFiles = ls.includes(unusedFilesName)
+    ? fs
+        // tslint:disable-next-line:non-literal-fs-path -- Not a reference to the fs package
+        .readFile(unusedFilesName)
+        .split(/\r?\n/g)
+        .filter(Boolean)
+    : [];
+  if (ls.includes(unusedFilesName) && !otherFiles.length) {
```
elibarzilay, line 280: It sounds a bit better to put these tests here since they become a harder policy, but regardless, it sounds better to keep related code together -- so maybe put here the relative paths test too?

jablko: Agreed, I put the relative paths check [here](https://github.com/microsoft/DefinitelyTyped-tools/pull/136/files#diff-926db8240fe6ae52f6ba3f2d7d343184e4772f3645551cda13d18748b9933acbR285) in #127.
- #127 added the relative paths check to @definitelytyped/definitions-parser
- DefinitelyTyped/dt-mergebot#200 removes the relative paths check from dt-mergebot
- This (#136) adds the empty `OTHER_FILES.txt` check to @definitelytyped/definitions-parser
- TBD removes the empty `OTHER_FILES.txt` check from dt-mergebot

So I think we're of the same mind?

elibarzilay (approved):

sandersn:
@elibarzilay is this ready to merge? I was planning to ship a new version of DT-tools next week so this might be good to include.

elibarzilay:
@sandersn Yeah, sorry I didn't see it.

PR #138, opened 2020-10-19
Mark removed types as deprecated

jablko:
Loop over the `@types` npm packages and mark any that no longer exist in HEAD (either in the `types` directory or in `notNeededPackages.json`) as deprecated.


packages/publisher/src/publish-packages.ts
```diff
@@ -187,6 +192,85 @@ export default async function publishPackages(
...
+      limit: 250,
+      from
+    };
+    objects = await search("@types", opts);
+    for (const { name: fullNpmName } of objects) {
+      const name = fullNpmName.slice("@types/".length);
+      // If they don't exist in the types directory or in
+      // notNeededPackages.json then mark them deprecated. Reference the
+      // commit/pull request that removed them.
+      if (!allPackages.tryGetLatestVersion(name) && !allPackages.getNotNeededPackage(name)) {
```
sandersn, line 210: probably don't want to waste time deprecating already-deprecated things either.

packages/publisher/src/publish-packages.ts
```diff
@@ -187,6 +192,85 @@ export default async function publishPackages(
...
+        if (dry) {
+          log(`(dry) Skip deprecate removed package ${fullNpmName}`);
+        } else {
+          log(`Deprecating ${fullNpmName} with message: ${deprecatedMessage}`);
+          await client.deprecate(fullNpmName, "", deprecatedMessage);
+        }
+      }
+    }
+    from += objects.length;
+  } while (objects.length >= 250 && from <= 5000);
```
sandersn, line 272: what is 5000? I thought there were close to 7000 `@types` packages.

sandersn (changes requested):
Couple of requests:

1. I feel like this should be a periodic cleanup task rather than something that runs every publish. So maybe this should be a standalone scripty package like retag, or even a script in DT proper.
2. I'm curious about the list of affected packages and their deprecation messages. Can you dump those here? I want to know what that looks like before taking this change.

@orta can you review the graphql? I think you know it the best of any of us on the team. (It looks plausible to me though.)

packages/publisher/src/publish-packages.ts
jablko: I think the query takes care of this: It doesn't return deprecated results.

packages/publisher/src/publish-packages.ts
jablko: Yes, so unfortunately the best I've been able to do so far is loop over the first 5250 `@types` packages ... I hope it can be built upon and that it's useful in the meantime ...

The only way I've found to loop over npm packages is the [registry API](https://github.com/npm/registry/blob/master/docs/REGISTRY-API.md#get-v1search) and it clamps `limit` at 250 and `from` at 5000. Without the `from <= 5000` condition it just repeats results 5000-5249 forever. The condition can be > 5000, it just repeats unnecessarily, and < 5000 gets fewer distinct results. So I think it's necessary and 5000 is optimal.

I hunted around for alternatives:
- I tried the [npms API](https://api-docs.npms.io/#api-Search) but the results I got were out of date. And it has a max of 5000 anyway.
- I looked for a dump of registry package names or a [public dataset](https://azure.microsoft.com/en-ca/services/open-datasets/catalog/?q=npm). No luck.
- I considered changing the sort parameters to e.g. get the last 5000 results plus the first 5000, which would cover all the `@types` packages of today, but think it's inelegant and ineffective: The registry API supports weighting results differently but not flipping the order entirely.

So [libnpmsearch](https://github.com/npm/libnpmsearch) is imperfect but it's fast, officially supported and has non-zero utility (I think). If it runs periodically I hope it'll catch results while they're in the first 5000. And hopefully one day the limit will increase, or an improved API will debut, or someone will think of something else?

jablko:
> 1. I feel like this should be a periodic cleanup task rather than something that runs every publish. So maybe this should be a standalone scripty package like retag, or even a script in DT proper.

Done :heavy_check_mark:

> 2. I'm curious about the list of affected packages and their deprecation messages. Can you dump those here? I want to know what that looks like before taking this change.

Currently the following five:
```Shell
Deprecating fingerprintjs2
Deprecating @types/fingerprintjs2 with message: 🤖 Merge PR #48103 feat: add fingerprintjs__fingerprintjs types by @lb… (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/48103)
Deprecating openapi-v3
Deprecating @types/openapi-v3 with message: Removed OpenApi-v3 types (#46819) (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/46819)
Deprecating fingerprintjs
Deprecating @types/fingerprintjs with message: 🤖 Merge PR #48103 feat: add fingerprintjs__fingerprintjs types by @lb… (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/48103)
Deprecating markerclustererplus
Deprecating @types/markerclustererplus with message: Change folder name to google__markerclustererplus to match scoped pac… (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/42121)
Deprecating storybook__polymer
Deprecating @types/storybook__polymer with message: [storybook__polymer] Remove types (#48796) (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/48796)
```

andrewbranch:
@sandersn, would you mind updating your review?

I’m fine with this, but the package.json needs `private: true` so lerna doesn’t publish it, and there needs to be a README that says what it does.

sandersn (approved):

raghavthind2005 (approved):

packages/deprecate/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "@definitelytyped/deprecate",
+  "version": "1.0.0",
```
andrewbranch, line 3:
```suggestion
  "private": true,
  "version": "1.0.0",
```

andrewbranch (changes requested):

jablko:
@andrewbranch I dusted off this PR and rebased it on https://github.com/microsoft/DefinitelyTyped-tools/pull/455 and https://github.com/microsoft/DefinitelyTyped-tools/pull/453. The [dry run shows](https://github.com/microsoft/DefinitelyTyped-tools/runs/6374013001?check_suite_focus=true#step:7:13) the deprecating it will do, currently.

PR #141, opened 2020-10-27
Add @tsconfig/recommended for easy extending

toinehartman:
[@tsconfig/recommended](https://github.com/tsconfig/bases#readme) provides a good basic `tsconfig.json` to extend.

### DIFF
--- a/packages/definitions-parser/allowedPackageJsonDependencies.txt
+++ b/packages/definitions-parser/allowedPackageJsonDependencies.txt
@@ -26,6 +26,7 @@
 @storybook/addons
 @storybook/react
 @testing-library/dom
+@tsconfig/recommended
 @tweenjs/tween.js
 @types/base-x
 @types/expect


andrewbranch:
I don’t think this makes sense for a DefinitelyTyped dependency? Can you explain what you need this for?

toinehartman:
> I don’t think this makes sense for a DefinitelyTyped dependency? Can you explain what you need this for?

It is part of my effort on [DefinitelyTyped#49166](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/49166), which I cannot get to work on all TSC versions without extending this TS config.

andrewbranch:
package.json dependencies in DefinitelyTyped are for pulling in _types_ that the published types will use. The tsconfig.json is not published to npm (there’s no reason for it to be), so this would result in @types/autoprefixer consumers needlessly downloading @tsconfig/recommended as an additional dependency.

> I cannot get to work on all TSC versions without extending this TS config.

First of all, there’s nothing magic about this config or `extends`—you can always replicate the effect of an `extends` by copying the relevant properties into your own config. However, DefinitelyTyped tsconfig files are restricted to a small set of changes to ensure that consumers can use the types under a wide range of their own config options, so modifying a bunch of options will be rejected anyway. If something is not compiling, it probably indicates an actual problem with the typings that needs to be solved.

If you need help with the PR, you can try asking the package owners (pinged by typescript-bot already); otherwise a DT maintainer (probably myself, next week) will take a look eventually. Thanks!

toinehartman:
Okay, I understand. Thanks for the thorough explanation. I tried to do what you suggest - basically copying the required `tsconfig.json` into mine - but that does not work. I will probably mention you in the other PR ([DefinitelyTyped#49166](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/49166)).

PR #142, opened 2020-10-28
Use libraryName homepage for not-needed stub

jablko:
Instead of tracking `sourceRepoURL` in `notNeededPackages.json` get it directly from the replacement npm package.

- Move `NotNeededPackage.readme` -> `generateNotNeededPackage()` unchanged except for the reference to `info.homepage` vs. `sourceRepoURL`
- Add `homepage` to `NpmInfo`
- Remove references to `sourceRepoURL`

### DIFF
--- a/packages/publisher/src/generate-packages.ts
+++ b/packages/publisher/src/generate-packages.ts
@@ -1,4 +1,5 @@
 import { makeTypesVersionsForPackageJson } from "@definitelytyped/header-parser";
+import assert = require("assert");
 import { emptyDir, mkdir, mkdirp, readFileSync } from "fs-extra";
 import * as path from "path";
 import * as yargs from "yargs";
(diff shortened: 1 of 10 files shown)


andrewbranch (approved):

packages/publisher/src/generate-packages.ts
```diff
@@ -241,7 +246,6 @@ export function createNotNeededPackageJSON(
     main: "",
     scripts: {},
     author: "",
-    repository: registry === Registry.NPM ? sourceRepoURL : "https://github.com/types/_definitelytypedmirror.git",
```
sandersn, line 244: I think this is deleted because both npm and github deprecated packages get the project url from npm. Is that right?

I'm not actually sure why github packages have a weird default. My best guess: I must have thought the url pointed to npm when I was writing the code.

sandersn (approved):
I had a question but I think I answered it myself.

PR #147, opened 2020-10-29
[definitions-parser] Fix deleting old versions of packages

### LINKED ISSUE #118: cannot remove old  DT version
I could be wrong, but it seems one cannot remove old version of the DT anymore.
In the case of DefinitelyTyped/DefinitelyTyped#47781 v4 of the package is to be removed (3 years old, just burden on CI) but it cannot get pass the check:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/47781/checks?check_run_id=1137596113
If I understand correctly, the version check picks v4 from the diff (but this is to be removed) and tries to match that version in the versions of current packages (current => 9 and 8):
```json
// part of parsed packages information
versions: [
    Semver { major: 9, minor: 0, patch: 0 },
    Semver { major: 8, minor: 1, patch: 0 }
  ],
```
```json
// part of affected versions information
[
  { name: 'fs-extra', version: '*' },
  { name: 'fs-extra', version: { major: 4, minor: undefined } }
]
```
as the version 4 is missing, the CI fails the check.
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/47781/checks?check_run_id=1137596113#step:5:27

andrewbranch:
Honestly, this `tryResolve` method seems extremely suspicious both before and after this change. It returns its input when resolution fails? I’m not sure I see the point. But, it seems to fix the bug, so the enigmatic `AllPackages` will live unscathed for another day, I suppose.

Fixes #118.

### DIFF
--- a/packages/definitions-parser/src/packages.ts
+++ b/packages/definitions-parser/src/packages.ts
@@ -63,7 +63,7 @@ export class AllPackages {
 
   tryResolve(dep: PackageId): PackageId {
     const versions = this.data.get(getMangledNameForScopedPackage(dep.name));
-    return versions ? versions.get(dep.version).id : dep;
+    return (versions && versions.tryGet(dep.version)?.id) || dep;
   }
 
   /** Gets the latest version of a package. E.g. getLatest(node v6) was node v10 (before node v11 came out). */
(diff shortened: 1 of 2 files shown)


sandersn (approved):

peterblazejewicz:
decent sample of code exegesis! 👍

PR #150, opened 2020-10-30
[perf] Fix versions showing up as "*" all over the place

andrewbranch:
I messed up a bunch of string <---> object version conversions in the perf bot code when trying to adapt to @alloy’s changes that allowed minor version typings directories. You can see a symptom of that mistake [here](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/48971#issuecomment-719085291), where both versions of react-dom that are tested are reported as `react-dom/*`. This fix allows `*` to still be used as a version _input_ on the CLI (e.g. `@definitelytyped/perf benchmark react-dom/*`), but should always resolve the `*` to the version number encoded in the header by the time it gets used as outputs (e.g. bot comments or database records).

### DIFF
--- a/packages/definitions-parser/src/lib/definition-parser.ts
+++ b/packages/definitions-parser/src/lib/definition-parser.ts
@@ -154,6 +154,32 @@ export function parseVersionFromDirectoryName(directoryName: string | undefined)
   };
 }
 
+/**
+ * Like `parseVersionFromDirectoryName`, but the leading 'v' is optional,
+ * and falls back to '*' if the input format is not parseable.
+ */
+export function tryParsePackageVersion(versionString: string | undefined): DependencyVersion {
+  const match = /^v?(\d+)(\.(\d+))?$/.exec(versionString!);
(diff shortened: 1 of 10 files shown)


packages/definitions-parser/src/lib/definition-parser.ts
```diff
@@ -154,6 +154,32 @@ export function parseVersionFromDirectoryName(directoryName: string | undefined)
   };
 }
 
+/**
+ * Like `parseVersionFromDirectoryName`, but the leading 'v' is optional,
```
andrewbranch, line 158: The root cause of a lot of version numbers getting lost was plugging a string like `"17"` into this function, where I was expecting to get back `{ major: 17, minor: undefined }`, but instead got back `undefined` because this function expects the format `v17`, since it was written only for directory names.

orta (approved):
Nice, good explanation 👍🏻

alloy (approved):
Thanks 😬

PR #156, opened 2020-11-05
Allow @maxim_mazurok/gapi.client.sheets in DT

Maxim-Mazurok:
Allow @maxim_mazurok/gapi.client.sheets in DT instead of gapi.client.sheets
Related to "A file cannot have a reference to itself." error in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/49347


andrewbranch:
Ah, because of `baseUrl`. Interesting.

PR #159, opened 2020-11-06
Allow tests to path-reference other tests

jablko:
Allows tests to use `/// <reference path="ts1.0/mock-tests.ts" />` to reference older-`typesVersions` tests, while still forbidding them to path-reference published files.

`if (ref.endswith(".d.ts"))` comes from :point_down: where it already distinguishes files that will be published (types) from others (tests)
https://github.com/microsoft/DefinitelyTyped-tools/blob/1a1420912ee1ede0c9f4e191ed4ffdc4ea0fb1a7/packages/definitions-parser/src/lib/module-info.ts#L169-L173

### DIFF
--- a/packages/definitions-parser/src/lib/module-info.ts
+++ b/packages/definitions-parser/src/lib/module-info.ts
@@ -387,7 +387,9 @@ export function getTestDependencies(
     let referencesSelf = false;
 
     for (const { fileName: ref } of referencedFiles) {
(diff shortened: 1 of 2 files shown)


andrewbranch:
This seems fine, but what exactly is the motivation for it?

jakebailey:
This whole thing has been replaced with a new lint, now that `module-info` is gone.

PR #160, opened 2020-11-07
Allow typesVersions-parent tests in tsconfig.json

jablko:
Allows e.g. listing `../mock-tests.ts` in `types/mock/ts1.0/tsconfig.json`, so you can run the same tests on newer and older `typesVersions`.

### DIFF
--- a/packages/definitions-parser/src/lib/definition-parser.ts
+++ b/packages/definitions-parser/src/lib/definition-parser.ts
@@ -411,7 +411,7 @@ function checkFilesFromTsConfig(packageName: string, tsconfig: TsConfig, directo
     if (file.startsWith("./")) {
       throw new Error(`In ${tsconfigPath}: Unnecessary "./" at the start of ${file}`);
     }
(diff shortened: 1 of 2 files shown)


packages/definitions-parser/src/lib/definition-parser.ts
```diff
@@ -411,15 +411,15 @@ function checkFilesFromTsConfig(packageName: string, tsconfig: TsConfig, directo
     if (file.startsWith("./")) {
       throw new Error(`In ${tsconfigPath}: Unnecessary "./" at the start of ${file}`);
     }
-    if (!isRelativePath(file)) {
+    if (!isRelativePath(file.replace(/^(?:\.\.\/)+/, ""))) {
```
andrewbranch, line 414: Doesn’t this allow tests from _anywhere_? Or is there something else that would prevent `types/react-dom/tsconfig.json` from listing `../react/react-tests.d.ts`?

PR #163, opened 2020-11-10
Update TS version tags weekly

sandersn:
Probably not quite right, but this should be a good start.

### DIFF
--- a/.github/workflows/update-ts-version-tags.yml
+++ b/.github/workflows/update-ts-version-tags.yml
@@ -3,7 +3,6 @@ name: Update ts* tags for ATA
 # For testing
 on: pull_request
 
-
 # For production
 # on:
 #   schedule:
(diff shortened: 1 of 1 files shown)


sandersn:
I left in `on: push` in order to test this, but I don't actually think it runs on PRs. I'll poke around and see whether I can set that up, or if I need to merge and just push commits to master.

sandersn:
Changed to `on: pull_request`, which allows me to test, and switched to a different NPM token, which has the correct rights for tagging `@types` packages.

That's all working; I'm not sure yet whether I've got the github token correct.

sandersn:
Everything is working except for github tagging; I haven't got GH_API_TOKEN correct. I think that's good enough to merge, since the github mirror isn't that important post-npm-acquisition. But I'm going to try to get the right token for github tagging before merging since it's a lot easier to test on push.

sandersn:
OK, I got the github token right. I'm going to let the github retag finish before switching to a scheduled run and then merging.

sandersn:
Looks like some versions of some packages are missing in the github mirror. I'm going to merge this and fix up the missing versions on github manually.

PR #169, opened 2020-11-12
Add republish manual script

sandersn:
Also update READMEs with some usage, as well as to reflect the now-automated ATA retagging.


andrewbranch (approved):
I think the name of the script should reflect that it only publishes to GitHub, but implementation looks good.

PR #171, opened 2020-11-13
Remove dependency from deprecated packages

sandersn:
1. This is only useful if you install `@types/X` but forget to install `X`, which is unlikely.
2. If you already have `X`, it does nothing.
3. A `"*"` dependency in `@types/X` apparently prevents unpublishing `X`.

### DIFF
--- a/packages/publisher/src/generate-packages.ts
+++ b/packages/publisher/src/generate-packages.ts
@@ -242,7 +242,7 @@ export function createNotNeededPackageJSON(
     scripts: {},
     author: "",
     repository: registry === Registry.NPM ? sourceRepoURL : "https://github.com/types/_definitelytypedmirror.git",
-    license,
+    license
   };
   if (registry === Registry.Github) {
     (out as any).publishConfig = { registry: "https://npm.pkg.github.com/" };
--- a/packages/publisher/src/generate-packages.ts
+++ b/packages/publisher/src/generate-packages.ts
@@ -243,10 +243,6 @@ export function createNotNeededPackageJSON(
     author: "",
     repository: registry === Registry.NPM ? sourceRepoURL : "https://github.com/types/_definitelytypedmirror.git",
     license,
-    // No `typings`, that's provided by the dependency.
-    dependencies: {
-      [libraryName]: "*"
-    }
   };
   if (registry === Registry.Github) {
     (out as any).publishConfig = { registry: "https://npm.pkg.github.com/" };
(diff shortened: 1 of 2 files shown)


andrewbranch (approved):
I guess that makes sense

sandersn:
Ryan points out that (1) may happen in the ATA case if you `require` a deprecated package with ATA support. We're investigating.

sandersn:
Nope, publication removes entries from types-registry that have been added to notNeededPackages.json

RyanCavanaugh:
If you `npm install @types/foo`, you might be pulling `foo` from a CDN and not need it locally.

The originating claim here is basically invalid; if you publish something to npm it is entirely normal in the ecosystem for someone to take a dependency on it. npm support can help you unpublish in exception circumstances but it is not normal to regularly unpublish things and this shouldn't be happening with any frequency. I don't find the OP compelling at all and don't see why we should avoid taking legit dependencies. Don't publish things to the public npm registry if you're not OK with someone taking a dependency on it.

sandersn:
So, to summarise:

1. The dependency is **rarely** harmful -- when somebody wants to unpublish a package with a matching, deprecated `@types` package.
2. The dependency is **rarely** helpful -- when somebody `npm i --save-dev` a deprecated `@types` package, but intend to get the actual package from a CDN. This saves them from confusion and having to also add the actual package as a devDependency.
3. The effort to remove the dependency from future deprecated packages is low.

We could make a call on whether to take this change by estimating the likelihood of (2) and finding it to be near-zero (say, by sampling the deprecated package list and checking how many usage instructions mention CDNs), or by seeing (1) happen again. For now, I'm going to close the PR.

andrewbranch:
I also don’t think the unpublishability reason has any weight; I’m only considering how the dependency is supposed to benefit the user and how else it might be harmful. We could change this without removing the dependency altogether, but the `*` version is pretty misleading _unless_ the user has the source library installed. If someone was pulling `foo` v2 from a CDN, and so installed `@types/foo@2` but no `foo` via npm, they’re going to end up using the latest version (say 3.0.0) of `foo` for typings but 2.0.0 at runtime. So I think the one place we’ve identified where the dependency might be helpful is actually a double-edged sword.

andrewbranch:
That said, leaving things the way they are now when the only actual complaint was not based on a good reason seems fine.

yGuy:
I would argue that my actual complaint was based on a good reason:

Some random guy on the internet "accidentally" published a commercial, closed-source software on npm, publicly. This package did not even exist on npm before he published it. However there was this @types package that depended on that software by name using the "*" syntax, three years old, *then empty* (because the original private package has typings built-in for years) and thus deprecated. That guy realized their mistake and the license breach and tried to "unpublish" almost immediately. But it took 36 hours to remove the package from the registry and this only worked through a complete DMCA take-down of the package because of the npm bug (and the dependency in @types). With the proposed changes in place, the package could have been removed within minutes and the damage would have been minimal.

PR #188, opened 2021-01-04
Don’t cache allowedPackageJsonDependencies indefinitely

andrewbranch:


andrewbranch:
Merging because publisher is down and I don’t think anyone is here to review

PR #196, opened 2021-01-21
Add manual timeout to downloadAndExtractFile

sandersn:
Open question:
I'm not sure whether to add a clearTimeout call to `on("error", reject)` -- right now I only put them in things that resemble normal returns.


packages/utils/src/io.ts
```diff
@@ -219,6 +223,7 @@ export function downloadAndExtractFile(url: string, log: LoggerWithErrors): Prom
         extract.on("error", reject);
```
andrewbranch, line 223: Ideally need a `rejectAndClearTimeout` function for here and line 232

andrewbranch (approved):

PR #202, opened 2021-02-10
Add 4.3 to supported TypeScript versions.

DanielRosenwasser:


sandersn (approved):
The change looks correct, as long as tests pass then it should be good to go.

PR #204, opened 2021-02-17
fix failure due to local changes to package.json

DetachHead:
fix "Your local changes to the following files would be overwritten by checkout" error caused when installing dependencies modifying package.json

see:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/44671#issuecomment-642138999
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/51262


andrewbranch (approved):
Good catch!

PR #205, opened 2021-02-17
force checkout

DetachHead:
the change I made in https://github.com/microsoft/DefinitelyTyped-tools/pull/204 didn't work as I missed the other checkout command

See https://dev.azure.com/definitelytyped/DefinitelyTyped/_build/results?buildId=87395&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=40b1ee41-44d6-5bba-aa04-4b76a5c732e5

Strangely, as far as I can tell it was failing on the second checkout before #204, but now it's failing on the first checkout instead

Not quite sure how to test this tbh


andrewbranch (approved):
Works for me

PR #206, opened 2021-02-19
Add timeout to https.request.

sandersn:
I haven't looked at the source, but Stack Overflow claims there is no timeout by default.

### DIFF
--- a/packages/utils/src/io.ts
+++ b/packages/utils/src/io.ts
@@ -100,6 +100,7 @@ export interface FetchOptions {
   readonly path: string;
   readonly retries?: boolean | number;
   readonly body?: string;
+  readonly timeout?: number;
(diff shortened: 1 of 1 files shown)


packages/utils/src/io.ts
```diff
@@ -100,6 +100,7 @@ export interface FetchOptions {
   readonly path: string;
```
sandersn, line 100: I'm not clear on why we maintain our own options object here instead of, at most, intersecting with http.RequestOptions.

andrewbranch: I can’t think of any reason. This seems bad.

sandersn: I'll fix it.

andrewbranch (approved):

PR #208, opened 2021-02-26
fix: parser fails with older scoped packages.

fedemp:
Given a path mapping as

```
"paths": {
	"@ckeditor/ckeditor5-utils": [
		"ckeditor__ckeditor5-utils/v10"
	]
}
```

definitions-parser failed to understand that `/v10` is not part of the
package name, and that the key of the mapping must be a scoped name.

This PR would unblock https://github.com/DefinitelyTyped/DefinitelyTyped/pull/51335

### DIFF
--- a/packages/utils/src/miscellany.ts
+++ b/packages/utils/src/miscellany.ts
@@ -30,11 +30,8 @@ export function unmangleScopedPackage(packageName: string): string | undefined {
   return packageName.includes(separator) ? `@${packageName.replace(separator, "/")}` : undefined;
 }
 
-export function removeVersionFromPackageName(packageName:string | undefined): string | undefined {
-	const firstSlash = packageName?.indexOf("/");
-	const lastSlash = packageName?.lastIndexOf("/");
-
-	return packageName && firstSlash !== lastSlash ? packageName.slice(firstSlash, lastSlash) : packageName;
+export function removeVersionFromPackageName(packageName: string | undefined): string | undefined {
+  return packageName?.replace(/\/v(\d){1,}$/i, "");
 }
 
 export async function sleep(seconds: number): Promise<void> {
--- a/packages/utils/src/miscellany.ts
+++ b/packages/utils/src/miscellany.ts
@@ -30,6 +30,13 @@ export function unmangleScopedPackage(packageName: string): string | undefined {
   return packageName.includes(separator) ? `@${packageName.replace(separator, "/")}` : undefined;
 }
 
+export function removeVersionFromPackageName(packageName:string | undefined): string | undefined {
+	const firstSlash = packageName?.indexOf("/");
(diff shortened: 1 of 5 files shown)


andrewbranch (commented):
Thank you! Would you mind adding a test?

peterblazejewicz:
@fedemp fedemp/DefinitelyTyped-tools/pull/1

fedemp:
@peterblazejewicz Thank you. I've merged your PR.

@andrewbranch Tests have been contributed.

andrewbranch:
Sorry, I should have been more specific. I have no doubt that `removeVersionFromPackageName` does what it says, but I was hoping to get a test for the actual logic change of the parser in here: https://github.com/microsoft/DefinitelyTyped-tools/blob/master/packages/definitions-parser/test/definition-parser.test.ts

fedemp:
> Sorry, I should have been more specific. I have no doubt that `removeVersionFromPackageName` does what it says, but I was hoping to get a test for the actual logic change of the parser in here: https://github.com/microsoft/DefinitelyTyped-tools/blob/master/packages/definitions-parser/test/definition-parser.test.ts

I've added a test, which allowed me to notice that there were still some issues. The problem is that checking that the typings are defined is the best I could come up with; there isn't any useful information from `getTypingInfo`.

https://github.com/microsoft/DefinitelyTyped-tools/pull/208/files#diff-1835dd7ec74428c084a5b69def3188bd2b7ec7d199efd2e0e83c10b9435254a7R68

andrewbranch:
I think that’s fine—it validates that the crash is gone, which is what we care about. I’m going to let the on-call DT person take a look and decide whether it’s an ok time to merge and publish. Thanks so much for taking the time to fix this!

fedemp:
Only thing missing is that I'm not sure if the publisher has to be modified too. :(

andrewbranch:
I suspect it doesn’t.

packages/utils/src/miscellany.ts
```diff
@@ -24,12 +24,25 @@ export function computeHash(content: string): string {
...
   return packageName.includes(separator) ? `@${packageName.replace(separator, "/")}` : undefined;
 }
 
+// Reverts unmangleScopedPackage.
+export function mangleScopedPackage(packageName: string): string {
+  return isScopedPackage(packageName) ? packageName.replace(/\//, "__").replace("@", "") : packageName;
+}
+
+export function removeVersionFromPackageName(packageName: string | undefined): string | undefined {
+  return packageName?.replace(/\/v(\d){1,}$/i, "");
```
elibarzilay, line 43: Looks like this should be simplified to `/\/v\d+$/i` --- the group is not needed, and the `{1,}` is equivalent to a `+`.

elibarzilay:
(Merging instead of waiting for a better regexp, but IMO it's worth improving...)

PR #209, opened 2021-02-28
feat: add @prisma/client to allowed package list

euxn23:


andrewbranch:
This is no longer needed, right?

euxn23:
No, this is needed.

andrewbranch:
I thought https://github.com/DefinitelyTyped/DefinitelyTyped/pull/51493 was meant to replace https://github.com/DefinitelyTyped/DefinitelyTyped/pull/51469?

euxn23:
After this p-r merged, I will add @prisma/client to type definition.

andrewbranch:
I don’t think `@prisma/client` is ever safe to be a DT dependency, because it depends on user-generated files in node_modules: https://unpkg.com/browse/@prisma/client@2.18.0/index.d.ts

euxn23:
Oh, I overlooked that. Now I also think this library is not safe to DefinitelyTyped, so I close this p-r. Thx.

PR #212, opened 2021-03-05
Fix old and same newer pkg require different TS

fedemp:
The `checkTypeScriptVersions` fails on an older package that depends on
a lower TS version than its latest version.

e.g. `@ckeditor/ckeditor-engine` depends on TS 3.6 and
`@ckeditor/ckeditor-engine/v11` depends on TS 2.3.

Without this fix, parser complains that v11 depends on a package that
has a more recent TS version as dependency.

One could argue that we could just drop the `// TypeScript Version` from
the older package, but it may occur that the older version requires
`LTS+1` and latest requires `LTS+2`.

### DIFF
--- a/packages/definitions-parser/src/check-parse-results.ts
+++ b/packages/definitions-parser/src/check-parse-results.ts
@@ -72,7 +72,7 @@ export async function checkParseResults(
 function checkTypeScriptVersions(allPackages: AllPackages): void {
   for (const pkg of allPackages.allTypings()) {
     for (const dep of allPackages.allDependencyTypings(pkg)) {
-      if (dep.minTypeScriptVersion > pkg.minTypeScriptVersion) {
+      if (dep.minTypeScriptVersion > pkg.minTypeScriptVersion && dep.name !== pkg.name) {
         throw new Error(`${pkg.desc} depends on ${dep.desc} but has a lower required TypeScript version.`);
       }
     }


andrewbranch:
Why does the newer version of the package appear as a dependency of the older? Is there a DT PR that shows the problem scenario?

fedemp:
There isn't any yet. I stumbled into this problem while typing `@ckeditor/ckeditor5-engine`.

Latest version requires TS 3.6 while old version suggest TS 2.3.

Maybe it has to do with scoped packages again, cause I've just noticed that `yargs` suggests different TS version for older versions, and even so, it works fine.

I guess I should close this until I get a more valid explanation of why it happened.

fedemp:
@andrewbranch I've created a PR for `ckeditor__ckeditor5-engine` and this error happened on CI.

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/51631/checks?check_run_id=2057065051

`ckeditor__ckeditor5-engine` only imports itself in the tests.

andrewbranch:
I could be wrong, but it feels to me like the dependency graph analysis itself is wrong. ckeditor5-engine/v11 does _not_ depend on ckeditor5-enging (latest), but it seems to be showing up in the dependencies here. It seems like that might be the actual bug here.

fedemp:
You are right. There was a problem with the mappings. I'll create a new PR.

PR #215, opened 2021-03-10
Fix path mapping not created for older scoped packages

fedemp:
Suppose we have a package called `@ckeditor/ckeditor5-utils` that has an older version `v10`.

The `tsconfig.json` file for the older version must include a path mapping to itself.''

```json
// tsconfig in v10
"paths": {
    "@ckeditor/ckeditor5-utils": ["ckeditor__ckeditor5-utils/v10"]
},
```

Currently `definition-parser` ignores any scoped package and does not try to create a path mapping for those.

### DIFF
--- a/packages/definitions-parser/src/lib/definition-parser.ts
+++ b/packages/definitions-parser/src/lib/definition-parser.ts
@@ -24,7 +24,9 @@ import {
   flatMap,
   unique,
   unmangleScopedPackage,
-  removeVersionFromPackageName
+  removeVersionFromPackageName,
+  hasVersionNumberInMapping,
+  mangleScopedPackage
 } from "@definitelytyped/utils";
 import { TypeScriptVersion } from "@definitelytyped/typescript-versions";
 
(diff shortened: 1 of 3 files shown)


packages/definitions-parser/src/lib/definition-parser.ts
```diff
@@ -473,7 +475,9 @@ function calculateDependencies(
       if (dependencyName !== scopedPackageName) {
         throw new Error(`Expected directory ${pathMapping} to be the path mapping for ${dependencyName}`);
       }
-      continue;
+      if (!hasVersionNumberInMapping(pathMapping)) {
```
fedemp, line 478:
```
            "@ckeditor/*": [
                "ckeditor__*"
            ]
```
should be ignored for path mappings.

fedemp:
Currently blocks https://github.com/DefinitelyTyped/DefinitelyTyped/pull/51631

andrewbranch (approved):
I think this is right? I’m ok with trying it.

PR #221, opened 2021-03-23
Move publisher to Azure Functions

andrewbranch:
First iteration:
- [x] Single timer trigger to replace long-running setInterval call
- [x] Ensure all logging is visible
- [x] Update deploy.yml
- [x] Ensure concurrent runs don’t happen by writing a lock file to blob storage

Later:
- Add HTTP trigger as webhook from DefinitelyTyped
- Instructions for running locally


sandersn (approved):

andrewbranch:
@sandersn I’m going to leave the test HTTP trigger in so we can trigger a run manually if needed. It requires a key as a query parameter to execute—you can get the full URL with the key from the Azure portal.

PR #228, opened 2021-04-03
add tests for typingsdata

43081j:
im not sure how useful these are, but i was looking around the repo to get some understanding of it and saw they weren't already covered by tests.

interestingly i noticed `fullNpmName` transforms scoped names to mangled names, but `fullEscapedNpmName` does not. maybe they are meant to behave the same?

a name of `@foo/bar` would result in `fullNpmName` being `@types/foo__bar` but `fullEscapedNpmName` being `@types%2f@foo/bar`. though i have a feeling maybe the names of these packages wouldn't be scoped anyway?

also, `NotNeededPackage#globals` returns `this.globals`, but surely thats an infinite loop unless im missing something? is there some precedence magic going on?

### DIFF
--- a/packages/definitions-parser/test/packages.test.ts
+++ b/packages/definitions-parser/test/packages.test.ts
@@ -1,8 +1,10 @@
 import { createMockDT } from "../src/mocks";
 import { getTypingInfo } from "../src/lib/definition-parser";
-import { AllPackages, TypingsVersions } from "../src/packages";
+import { AllPackages, TypingsVersions, TypingsData, License } from "../src/packages";
 import { parseDefinitions } from "../src/parse-definitions";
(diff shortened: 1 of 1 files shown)


andrewbranch (approved):
> interestingly i noticed `fullNpmName` transforms scoped names to mangled names....  though i have a feeling maybe the names of these packages wouldn't be scoped anyway?

That’s correct, `this.name` is already mangled. It looks like `fullNpmName()` doesn’t need to try to mangle anything. This could be cleaned up, but is currently working.

> also, `NotNeededPackage#globals` returns `this.globals`, but surely thats an infinite loop unless im missing something?

Yeah, I guess it’s never accessed. That can be removed.

PR #231, opened 2021-04-08
Fix: Allow /* in scoped versioned mappings.

fedemp:
* definition-parser: Use scoped names when testing mappings for relative
imports.
* miscellany/utils: When removing version number, keeps wildcard at the
end of mapping.
* miscellany/utils: When checking the existence of a version number,
catch those packages that include a wildcard path mapping.

### DIFF
--- a/packages/definitions-parser/test/definition-parser.test.ts
+++ b/packages/definitions-parser/test/definition-parser.test.ts
@@ -24,7 +24,7 @@ describe(getTypingInfo, () => {
       "index.d.ts",
       `// Type definitions for @ckeditor/ckeditor-engine 25.0
 // Project: https://github.com/ckeditor/ckeditor5/tree/master/packages/ckeditor5-engine
-// Definitions by: Federico <https://github.com/fedemp>
+// Definitions by: My Self <https://github.com/ñ>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 import * as utils from '@ckeditor/ckeditor5-utils';`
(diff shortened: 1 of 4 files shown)


packages/definitions-parser/test/definition-parser.test.ts
```diff
@@ -24,7 +24,7 @@ describe(getTypingInfo, () => {
       "index.d.ts",
       `// Type definitions for @ckeditor/ckeditor-engine 25.0
 // Project: https://github.com/ckeditor/ckeditor5/tree/master/packages/ckeditor5-engine
-// Definitions by: Federico <https://github.com/fedemp>
+// Definitions by: My Self <https://github.com/ñ>
```
fedemp, line 27: I'm removing mentions of **myself** cause I found it too egocentric.
`ñ` is not a valid github username, so there's no promotion either.

packages/utils/src/miscellany.ts
```diff
@@ -40,11 +40,11 @@ export function mangleScopedPackage(packageName: string): string {
 }
 
 export function removeVersionFromPackageName(packageName: string | undefined): string | undefined {
-  return packageName?.replace(/\/v(\d){1,}$/i, "");
+  return packageName?.replace(/\/v\d+(\/\*)?$/, "$1");
```
fedemp, line 43: Remove the case insensitive flag cause version directories must start with a lowercase `v`.

andrewbranch:
Is this a fix for [this](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/51546/files/9cd554f57ae8740db90842f8d4c87e84de202214#diff-88cee239d43150c8d6db3742f0ead49122251de4de038a6462edee731e661135)? I was literally just coming to do this.

andrewbranch (approved):

andrewbranch:
Thanks a lot!

fedemp:
> Is this a fix for [this](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/51546/files/9cd554f57ae8740db90842f8d4c87e84de202214#diff-88cee239d43150c8d6db3742f0ead49122251de4de038a6462edee731e661135)? I was literally just coming to do this.

It seems it is the same case. :)

PR #247, opened 2021-05-06
Actually log crashes to STDERR, replace trackEvent with trackException

andrewbranch:
`console.error` will allow error messages to show up in App Insights `traces`. I couldn’t find where this `trackEvent` shows up. I’m hoping and assuming that `trackException` will show up in `exceptions`, but the most important thing is having it in logs.


sandersn:
What does it mean to move `await removeLock()` after the logging?

andrewbranch:
Just in case something goes wrong with the file system operation, we still get an error logged.

PR #250, opened 2021-05-10
Add TS 4.4 as typescript@next

sandersn:


sandersn:
Probably should wait until 4.4 is available as `typescript@next` before shipping, although DT should be fine either way.

PR #252, opened 2021-05-12
definition-parser: Show path of empty dir

fedemp:
When checkAllUsedRecur fails because of an empty directory, show the
full path to it. Currently "[Object object]" is printed to the terminal.

debugPath does not show path relative to current position, but relative
to root of project.


andrewbranch (approved):
Thanks!

PR #253, opened 2021-05-12
Allow deleted files outside notNeededPackages entries

sandersn:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/52719 needs to deprecate @types/socket.io@2 and in doing so modify its dependents. One of its dependents needs to delete a file in its own package (its socket.io test is no longer needed), but this isn't allowed currently.

This changes it to be allowed. I'm not sure I've covered all the interesting cases, so check whether the new code is now too permissive.


andrewbranch (approved):

andrewbranch:
Merge and deploy at will

PR #254, opened 2021-05-14
Make lockfile expire after function timeout

andrewbranch:
This prevents Azure Functions from thinking there’s still an instance running if an old instance got unceremoniously terminated without being able to clean up after itself by assuming that an instance cannot still be running if it claimed to have started longer ago than the function timeout would allow. I’ve seen this happen exactly one time since moving to Azure Functions, so I think we can expect it to happen every once in a while, I guess.


packages/publisher/src/main.ts
```diff
@@ -44,28 +44,32 @@ export default function main() {
...
-      }
-    );
+  await writeFile(lockFilePath, JSON.stringify({ timestamp: currentTimeStamp() }));
+  cb().then(
+    () => remove(lockFilePath),
+    async error => {
+      console.error(error?.stack || error?.message || error);
+      applicationinsights.defaultClient.trackException({
+        exception: error,
+      });
```
andrewbranch, line 63: Ignore whitespace to review

packages/publisher/tsconfig.json
```diff
@@ -2,11 +2,7 @@
     "extends": "../../tsconfig.base.json",
     "compilerOptions": {
         "outDir": "dist",
-        "rootDir": "src",
-        "baseUrl": ".",
-        "paths": {
-            "*": ["src/types/*"]
-        }
+        "rootDir": "src"
```
andrewbranch, line 5: Drive by clean up 🚗

packages/utils/src/io.ts
```diff
@@ -39,19 +39,29 @@ export function readFileSync(path: string): string {
...
   try {
-    return await readJson(filePath);
+    return await readJson(filePath, isObject);
   } catch (e) {
     console.error(`Run ${generatedBy} first!`);
     throw e;
   }
 }
 
-export function readJsonSync(path: string): object {
```
andrewbranch, line 49: Pro tip, `JSON.parse` does not always produce an `object` 🤦 

Quality of life enhancements throughout this section

sandersn (approved):
Looks reasonable enough to try shipping next week!

PR #258, opened 2021-05-17
chore(node): readable-stream compat

SimonSchick:
We need to de-couple `readable-stream` from current head of @types/node since readable-stream strictly states it's a [copy of the v10 stream API](https://www.npmjs.com/package/readable-stream).

See corresponding DT PR: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/52732/files#diff-af7da952bbd3d5c94b761da7f4ce4b7b5dac398dd34b890bde3c77719d5191d6R4


andrewbranch:
@sandersn maybe dependencies on `@types/*` should be automatically allowed as long as the version specified is not what you’d get with an implicit reference.

SimonSchick:
I agree with the above.

PR #261, opened 2021-05-20
If the module is a single dts file, show it in the README

orta:
The idea being that you won't need to pop to GitHub to read the types if it's a small dep.

### DIFF
--- a/packages/publisher/src/generate-packages.ts
+++ b/packages/publisher/src/generate-packages.ts
@@ -277,10 +277,11 @@ export function createReadme(typing: TypingsData, packageFS: FS): string {
 
   if (typing.dtsFiles.length === 1 && packageFS.readFile(typing.dtsFiles[0]).length < 2500) {
     const dts = typing.dtsFiles[0]
-    lines.push(`## ${typing.dtsFiles[0]}`)
-    lines.push("```ts")
+    const url = `${definitelyTypedURL}/tree/${sourceBranch}/types/${typing.subDirectoryPath}/${dts}`
+    lines.push(`## [${typing.dtsFiles[0]}](${url})`)
+    lines.push("````ts")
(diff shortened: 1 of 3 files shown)


andrewbranch:
Keep in mind this doubles the unpacked size of the package—certainly less than double gzipped, but a handful of these definitions are huge. There should definitely be a size limit. Alternatively, or in addition to this, you could add links to unpkg and/or GitHub.

orta:
Agree, let's cap it at ~300 lines of code?

orta:
I'm gonna go with 2500 chars, that's pretty much 50% of all DT files, - here's one at 2497 https://github.com/DefinitelyTyped/DefinitelyTyped/blob/61b08d3df7ed2b878127ab0b73fa16f057a178df/types/react-imgpro/index.d.ts

andrewbranch:
Works for me 👍

packages/publisher/src/generate-packages.ts
```diff
@@ -275,6 +275,14 @@ export function createReadme(typing: TypingsData): string {
   lines.push("# Details");
   lines.push(`Files were exported from ${definitelyTypedURL}/tree/${sourceBranch}/types/${typing.subDirectoryPath}.`);
 
+  if (typing.dtsFiles.length === 1 && packageFS.readFile(typing.dtsFiles[0]).length < 2500) {
+    const dts = typing.dtsFiles[0]
+    lines.push(`## ${typing.dtsFiles[0]}`)
+    lines.push("```ts")
+    lines.push(packageFS.readFile(dts))
```
andrewbranch, line 282: Technically breakable by multiline comments, but not worth guarding against 😆 

````ts
/**
 * @example
```
gotcha()!
```
 */
declare function gotcha(): void;
````

andrewbranch (approved):

PR #268, opened 2021-05-27
Require lowercase package names in notNeededPackages.json

sandersn:
This means we can't deprecate packages with uppercase letters, but I don't think there are many on npm -- possibly none.

### DIFF
--- a/packages/definitions-parser/src/mocks.ts
+++ b/packages/definitions-parser/src/mocks.ts
@@ -12,7 +12,7 @@ class DTMock {
       JSON.stringify({
         packages: {
           angular: {
-            libraryName: "angular 2",
+            libraryName: "angular",
             asOfVersion: "1.2.3"
           }
         }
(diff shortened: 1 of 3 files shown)


packages/definitions-parser/src/mocks.ts
```diff
@@ -12,7 +12,7 @@ class DTMock {
       JSON.stringify({
         packages: {
           angular: {
-            libraryName: "Angular 2",
+            libraryName: "angular 2",
```
andrewbranch, line 15: This is also not a valid library name though?

sandersn: Yeah...I think at one point non-npm libraries were allowed here, but checking notNeededPackages, that's not the case. I'll adjust the test.

sandersn: oh, in fact there is now a check that both the types package and the libraryName package exist, so this mock data was just wrong.

PR #270, opened 2021-05-28
Allow node/buffer path mapping

sandersn:
This will allow https://github.com/DefinitelyTyped/DefinitelyTyped/pull/53441 to pass.


SimonSchick (approved):

andrewbranch (approved):

SimonSchick:
On that note, wouldn't it make sense to prioritize node modules over modules that happen to have the same name?

This might also be solved by eventually using the `node:buffer` module name.

PR #273, opened 2021-06-01
Add logging to withFileLock

sandersn:
1. On entrance
2a. On detecting a lock file
2b. On not detecting a lock file


sandersn:
Based on an outage from 3:30 AM - 1:30 PM PDT 2021-06-01 in which lock.json didn't get deleted after the end of a run.

andrewbranch (approved):
After fixing merge conflicts, the message `Lock file does not exist; writing lock file and running.` is logged even when the lockfile does exist but is expired, indicating a previous failure to delete it. That seems ok for now.

PR #275, opened 2021-06-12
Fix: remove minor version number from package name

fedemp:
Add support for minor versions when removing the version suffix from a
package name. E.g. @foo/bar@v27.1 should be @foo/bar, but currently the
regexp does not match the dot character.

### DIFF
--- a/packages/utils/src/miscellany.ts
+++ b/packages/utils/src/miscellany.ts
@@ -61,11 +61,11 @@ export function mangleScopedPackage(packageName: string): string {
 }
 
 export function removeVersionFromPackageName(packageName: string | undefined): string | undefined {
-  return packageName?.replace(/\/v\d+(\/\*)?$/, "$1");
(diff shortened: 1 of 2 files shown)


sandersn:
This makes sense, but where did you encounter this problem? Is there a PR on Definitely Typed?

fedemp:
> This makes sense, but where did you encounter this problem? Is there a PR on Definitely Typed?

Not yet. I have those changes locally. But take for example @ckeditor/ckeditor5-essentials which current latest version is 27.1 and there are breaking changes for version 28.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/131f18d3e007e537e0c8249ac85073d970042f1c/types/ckeditor__ckeditor5-essentials/index.d.ts

fedemp:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/53834 is blocked by this.

andrewbranch (approved):

PR #276, opened 2021-06-13
Added @types/eslint & co to allowedPackageJsonDependencies.txt

JoshuaKGoldberg:
Corresponds to https://github.com/DefinitelyTyped/DefinitelyTyped/pull/53803.


sandersn:
I should have commented here too, but I don't think these are needed: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/53803#pullrequestreview-683222305

PR #281, opened 2021-06-17
add buffer to allowed package json dependencies

TimoGlastra:
the `buffer` npm package ships its own types

### DIFF
--- a/packages/definitions-parser/allowedPackageJsonDependencies.txt
+++ b/packages/definitions-parser/allowedPackageJsonDependencies.txt
@@ -313,6 +313,7 @@ bookshelf
 boxen
 broadcast-channel
 bson
+buffer
 bullmq
 cac
 cash-dom


andrewbranch:
What happens if someone manages to end up with both `buffer` and `@types/node` in their node_modules? Which one wins, or are there errors? I _think_ this will work ok on DefinitelyTyped but want to get feedback from @sandersn first.

TimoGlastra:
That's what we have in our code base at the moment. 

If I do the following, TS will see it as importing the NodeJS buffer (because in NodeJS thats what would happen). But if the NodeJS globals are not available, TS will see it as importing from the NPM package. This allows us to use the NPM package in browsers/react native and the built-in buffer from NodeJS in node environments. So I think this wouldn't interfere with anything as the `buffer` NPM package does not have global types.

```
import { Buffer } from 'buffer'
```

If I do the following, TS will always see it as importing from the `buffer` npm package.

```
import { Buffer } 'buffer/'
```

andrewbranch:
Merging here, but we’ll want to pay special attention to the DT PR and make sure CI and publishing do the right thing.

PR #291, opened 2021-07-15
Add @types/mongodb to allowedPackageJsonDependencies

ImRodry:
I forgot to add this in #290 I'm sorry but this PR just adds `@types/mongodb` to the allowed deps list


andrewbranch:
Is this because some DT packages need to reference older versions of mongodb typings?

PR #295, opened 2021-07-27
Remove the audit CLI output

orta:
Lots of the CI output in when you use `dtslint-runner` is [noise](https://github.com/microsoft/TypeScript-DOM-lib-generator/runs/3174306610?check_suite_focus=true) - this should reduce that


andrewbranch (approved):
Thanks!

PR #296, opened 2021-08-02
chore: remove ts3.6 support

SimonSchick:
See https://github.com/DefinitelyTyped/DefinitelyTyped/pull/54895

Turns out this was harder than just 1 PR 😂

### DIFF
--- a/packages/header-parser/test/index.test.ts
+++ b/packages/header-parser/test/index.test.ts
@@ -40,7 +40,7 @@ describe("parse", () => {
       libraryName: "foo",
       libraryMajorVersion: 1,
       libraryMinorVersion: 2,
-      typeScriptVersion: "3.6",
+      typeScriptVersion: "3.7",
       nonNpm: false,
       projects: ["https://github.com/foo/foo", "https://foo.com"],
       contributors: [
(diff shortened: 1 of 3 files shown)


sandersn (changes requested):
I love that you're getting rid of node's 3.6 types, but I want to hold this particular PR until TS4.4 actually releases. Now that we've been on the same release cadence for two years, it makes sense to add support for a new version and remove support for an old version at the same time.

Let's keep this PR around and you or I can push the 4.4 release changes to it, then merge+ship it at the actual 4.4 release.

SimonSchick:
Yea np, would be nice if there was a script that does this automatically but I think it's simple enough.

sandersn:
I added 4.5 -- a bit late, since it was supposed to be when the RC released -- so when tests finish I'll merge this. Then I'll remove 3.6 from the non-node packages on DT later today.

SimonSchick:
🔥

sandersn (approved):

sandersn:
OK, it's in. Give me a few minutes to update dtslint and dts-critic and then DT's tests should be good to remove ts3.6/ directories.

sandersn:
OK, dtslint is up-to-date now, so DT now no longer tests on 3.6.

PR #306, opened 2021-08-30
Added new linting package

JoshuaKGoldberg:
Works on https://github.com/microsoft/dtslint/issues/300#issuecomment-904979821 by setting up a new `@definitelytyped/linting` package with the following utilities:

* `checkPackageJson `
* `checkTsconfig`
* `checkTslintJson`
* `testDependencies`

Once this package exists I'll be able to pull it into the scripts in DefinitelyTyped itself. That way DT won't need to rely on `dtslint` to call this logic.


JoshuaKGoldberg:
Closing as this has aged away. If you the reader would like this code, feel free to take it and send your own PR. ❤️

PR #309, opened 2021-09-05
Add @types/babel__core to allowed dependencies

SNDST00M:
Fix for DefinitelyTyped/DefinitelyTyped#55590.


sandersn:
As I commented on the PR, the fix you need is in tsconfig.json, not package.json.

PR #311, opened 2021-09-12
Allow new @maxim_mazurok/gapi.client.* packages

Maxim-Mazurok:
Adds new Google API Types packages into allowedPackageJsonDependencies.txt and removes old ones.

Related to https://github.com/Maxim-Mazurok/google-api-typings-generator/pull/554
As seen previously in #263 #240 #183 #175 #174 #156 #154
Blocks https://github.com/DefinitelyTyped/DefinitelyTyped/pull/55712

### DIFF
--- a/packages/definitions-parser/allowedPackageJsonDependencies.txt
+++ b/packages/definitions-parser/allowedPackageJsonDependencies.txt
@@ -89,6 +89,7 @@
 @maxim_mazurok/gapi.client.cloudtrace
 @maxim_mazurok/gapi.client.composer
 @maxim_mazurok/gapi.client.compute
+@maxim_mazurok/gapi.client.contactcenterinsights
 @maxim_mazurok/gapi.client.container
 @maxim_mazurok/gapi.client.containeranalysis
 @maxim_mazurok/gapi.client.content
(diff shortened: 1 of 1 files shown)


peterblazejewicz:
This broke DT, the line needs either to be reverted or relevant package should be updated on DT,
otherwise it blocks CI (and local tests):

> @maxim_mazurok/gapi.client.remotebuildexecution

peterblazejewicz:
https://github.com/DefinitelyTyped/DefinitelyTyped/runs/3590414786?check_suite_focus=true
```bash
Using local DefinitelyTyped at .
Parsing definitions...
Found 7682 packages.
Parsing in 2 child process(es)...
Error: In package.json: Dependency @maxim_mazurok/gapi.client.remotebuildexecution not in the allowed dependencies list.
If you are depending on another `@types` package, do *not* add it to a `package.json`. Path mapping should make the import work.
For namespaced dependencies you then have to add a `paths` mapping from `@namespace/*` to `namespace__*` in `tsconfig.json`.
If this is an external library that provides typings,  please make a pull request to microsoft/DefinitelyTyped-tools adding it to `packages/definitions-parser/allowedPackageJsonDependencies.txt`.
    at checkPackageJsonDependencies 
...
```

peterblazejewicz:
fixing #312

Maxim-Mazurok:
Hm, I didn't think about this, thanks for a quick fix!
I will make sure that the package is removed from DT first, before I remove it from here, cheers!

PR #312, opened 2021-09-13
Revert partially "gapi update (#311)"

peterblazejewicz:
This reverts part of commit 383abedd72df1df6a4dbf77b7a2a1e663de21b0a.

Adding back single package which is part of DT CI gate tests and fails due to this specific packae removal.
TBD: create proper entry on DT to handle this specific package updates

Thanks!


peterblazejewicz:
This needs review, if that is really a fix

andrewbranch:
Thanks @peterblazejewicz 👍

PR #314, opened 2021-09-14
add '@nick-thompson/elementary' to allowed package.json typings

Hrle97:
Here is the [DefinitelyTyped PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/55762).


sandersn (changes requested):
I looked at the DT PR. You do not need a dependency on `@nick-thompson/elementary` to provide types for it. So this PR is not needed.

PR #319, opened 2021-09-16
build(deps): adds d3 & neo4j-driver-lite

alma3lol:


andrewbranch:
Can you link the DT PR where this is needed?

alma3lol:
> Can you link the DT PR where this is needed?

I'm actually working on it. Should be ready by Monday. @andrewbranch

andrewbranch:
Post it here whenever it’s up and then we’ll review 👍

PR #327, opened 2021-10-05
Add `react-map-gl` to `allowedPackageJsonDependencies`

eps1lon:
See `types` field in https://unpkg.com/browse/react-map-gl@6.1.0/package.json


sandersn:
I'm confused about this. `react-map-gl` is itself on DT, and is currently failing because of a transitive dependency on gl-matrix. Is this PR supposed to help with that? Or with some dependent of react-map-gl?

eps1lon:
`react-map-gl` currently relies on implicit `children` (of e.g. `FunctionComponent`). Since they ship their own types now, we can just drop them from the DT repo.

sandersn (approved):

PR #332, opened 2021-10-10
Run npm install with --no-fund flag

imcotton:
To simplify logging during **dtslint** check from **DefinitelyTyped** repo.


imcotton:
Thanks for the suggestion, changed.

PR #336, opened 2021-10-17
Add @types/event-kit to allowed dependencies

jnv:
Hello, I would like to allow the new dependency for DefinitelyTyped/DefinitelyTyped#56580.

The type def for @atom/watcher package depends on Disposable from @types/event-kit,


sandersn (changes requested):
This is not needed. See my review on the DT PR.

PR #337, opened 2021-10-19
Add mts and cts support when we check for .d.ts file

xiaoxiangmoe:


andrewbranch:
Is that PR redundant with https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56495?

PR #341, opened 2021-11-03
Add `@types/formidable` to `allowedPackageJsonDependencies`

gboer:
See https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56925, so that we can safely upgrade `@types/formidable` without breaking other packages (`restify` specifically)


peterblazejewicz:
For me that is better approach, than asking folks to create v1 of `formidable` DT package (in v1 there were broken releases in DT, v2 in native package was crafted to address many aspects of `formidable`)

sandersn (approved):

PR #352, opened 2021-11-29
Automatically patch a package.json entry onto all object export maps

weswigham:
Some ecosystem packages unconditionally `require` a package's `package.json`, which throws if it's not exported in an export map. While a bundler has no good reason to be inspecting an `@types` package, so we really shouldn't need to be providing a `package.json` entrypoint, including the mapping should fix issues like [this one](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/57235#issuecomment-979829788).


weswigham:
See [this post](https://github.com/nodejs/node/issues/33460#issuecomment-948711481)

orta:
I agree, think this makes sense

andrewbranch (approved):

PR #364, opened 2021-12-03
Port dtslint tests to jest, re-root dts-critic tests

sandersn:
dts-critic tests need to be run from DefinitelyTyped-tools/ root now, which also changes a couple of the assertions.


sandersn:
tslint requires the rules to be built before running tests, but ts-jest doesn't build files first =(

package.json
```diff
@@ -9,7 +9,7 @@
     "pretest": "yarn build",
     "lint": "eslint . --ext .ts",
     "format": "prettier --write 'packages/**/*.ts'",
-    "test": "jest",
+    "test": "tsc -b packages/dtslint && jest",
```
andrewbranch, line 12: This should already be covered by `"pretest"` 🤔

sandersn: Hmm, makes me think that dtslint isn't correctly integrated into the project build.

sandersn: Yep, that was it.

PR #369, opened 2021-12-17
allow sass

alan-agius4:
This is needed to remove @types/sass

### DIFF
--- a/packages/definitions-parser/allowedPackageJsonDependencies.txt
+++ b/packages/definitions-parser/allowedPackageJsonDependencies.txt
@@ -501,6 +501,7 @@ remark
 rollup
 rrule
 rxjs
+sass
 safe-buffer
 scroll-behavior
 scroll-into-view-if-needed


sandersn:
Is there a PR for this on Definitely Typed? I looked and didn't see one.

alan-agius4:
@sandersn, I didn’t open a PR in DefinitelyTyped just yet, since the tests were failing locally.

But this is needed to remove `types/sass` since now `sass` includes its own declaration files. https://unpkg.com/browse/sass@1.45.0/

sandersn:
Looks like `@types/sass` has 4 dependencies inside DT. Are you able to update those to use the new sass types, or will they need to point to the old `@types/sass`? I'd prefer to add both at the same time to make future reference easier.

Adding @orta for next week in case this takes longer than today, since I'll be on vacation.

alan-agius4:
@sandersn, I opened a draft PR with the proposed changes https://github.com/DefinitelyTyped/DefinitelyTyped/pull/57768

sandersn (approved):

PR #374, opened 2022-01-04
Allow `@glimmer/interfaces` in DT types

chriskrycho:
Correctly supporting Ember v4 types takes a dependency on this package for a couple key APIs.


chriskrycho:
Note: this may not be the right approach here. I’m evaluating alternatives. Will update as I make progress.

chriskrycho:
Yeah, wrong approach, coming it at it a different way.

PR #385, opened 2022-01-19
Forbid adding packages in notNeededPackages.json

sandersn:
It leaves DT in an inconsistent state, which causes the publisher to assert.

### DIFF
--- a/packages/dtslint/src/index.ts
+++ b/packages/dtslint/src/index.ts
@@ -138,6 +138,7 @@ async function runTests(
     // so assert that we're really on DefinitelyTyped.
     assertPathIsInDefinitelyTyped(dirPath);
     assertPathIsNotBanned(dirPath);
+    assertPackageIsNotDeprecated(dirPath, await readFile("../notNeededPackages.json", "utf-8"));
   }
 
   const typesVersions = await mapDefinedAsync(await readdir(dirPath), async name => {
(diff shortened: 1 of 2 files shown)


andrewbranch (approved):

packages/dtslint/src/index.ts
```diff
@@ -138,6 +138,7 @@ async function runTests(
     // so assert that we're really on DefinitelyTyped.
     assertPathIsInDefinitelyTyped(dirPath);
     assertPathIsNotBanned(dirPath);
+    assertPackageIsNotDeprecated(dirPath, await readFile("../notNeededPackages.json", "utf-8"));
```
AviVahl, line 141: @andrewbranch I just cloned DT. trying to update a type package. When running `npm t <packageName>` I got:
```
Error: ENOENT: no such file or directory, open '../notNeededPackages.json'
```
Debugging revealed this to be the line that throws.

When calling `readFile`, if using a relative path as a parameter, it will use current cwd to resolve that to an absolute path. This might be right for specific cases where `notNeededPackage.json` is indeed just one level up the cwd, but breaks running dtslint from the root of DefinitelyTyped.

TNH, I'm not sure how there was no additional noise regarding this. Nobody executes tests locally? Or perhaps they are still running older dtslint? Or maybe this change took time to get released..?

plumdog: @AviVahl fwiw, I happen to have just hit the same thing. #402

sandersn: Thanks for the sleuthing @AviVahl. I'll see what I can do.

> Nobody executes tests locally? 

I would guess this is it.

AviVahl: Thanks for the fix :+1:

sandersn: Fix is up at #405

PR #403, opened 2022-02-16
Fix self-dependency versions

andrewbranch:
Ensures packages don’t depend on old versions of each other.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,10 +20,10 @@ jobs:
     - name: check dependencies
       run: |
         if cat yarn.lock | grep "@definitelytyped/"; then
-          echo "\
-            A package in this monorepo is depending on an older version of one of its siblings. \
-            Ensure all '@definitelytyped/*' dependencies in package.json files use the version \
-            listed in lerna.json to let lerna automatically keep them in sync."
+          echo "A package in this monorepo is depending on an older version of one of its siblings. \
+          Ensure all '@definitelytyped/*' dependencies in package.json files use the version \
+          listed in lerna.json to let lerna automatically keep them in sync."
+
           echo ""
           echo "yarn.lock"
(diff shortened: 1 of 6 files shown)


packages/dtslint/src/index.ts
```diff
@@ -20,6 +20,8 @@ async function main(): Promise<void> {
   let lookingForTsLocal = false;
   let tsLocal: string | undefined;
 
+  console.log(`dtslint@${require("../package.json").version}`);
```
sandersn, line 23: is this helpful long-term?

.github/workflows/ci.yml
```diff
@@ -17,6 +17,19 @@ jobs:
     - uses: actions/setup-node@v2
       with:
         node-version: ${{ matrix.node }}
+    - name: check dependencies
+      run: |
+        if cat yarn.lock | grep "@definitelytyped/"; then
```
sandersn, line 22: isn't this equivalent?
```suggestion
        if grep "@definitelytyped/" yarn.lock; then
```

I don't understand the scope of this command -- isn't this the root yarn.lock? How does this check the package yarn.locks? Or is yarn.lock a convenient single point where all incorrect dependencies show up?

sandersn (changes requested):
Couple of questions

.github/workflows/ci.yml
andrewbranch: For workspaces, there's a single yarn.lock.

And yes, I'm bad at bash.

packages/dtslint/src/index.ts
andrewbranch: I think so. I added it when the version mismatch was just a hunch, not yet confirmed, and not yet mitigated with the CI check. But I think it’s useful to be able to look at older logs and tell what was running.

sandersn (approved):

PR #404, opened 2022-02-16
Fix incorrect relative path to notNeededPackages.json

### LINKED ISSUE #402: Release 0.0.104 broke 'npm test [package]'
Clone https://github.com/DefinitelyTyped/DefinitelyTyped, cd into it and `npm i`, then:
```bash
$ npm i @definitelytyped/dtslint@0.0.103 && npm test through
...

> definitely-typed@0.0.3 test
> dtslint types "through"

$ npm i @definitelytyped/dtslint@0.0.104 && npm test through
...

> definitely-typed@0.0.3 test
> dtslint types "through"

Error: ENOENT: no such file or directory, open '../notNeededPackages.json'
```

I haven't tried to dig into why. It might be that I'm not understanding something, but I think I'm following the docs at https://github.com/DefinitelyTyped/DefinitelyTyped#running-tests.

plumdog:
Fixes #402 

I think this does fix this particular bug, but - assuming I have built everything correctly in my local env - I now get linting errors that I don't on 0.0.103:

Eg, `npm test through` now gives me a bunch of TypeError chatter, then:
```
ERROR: 3:5  prefer-const  Identifier 'i' is never reassigned; use 'const' instead of 'var'.
```

I don't know if this is expected.

### DIFF
--- a/packages/dtslint/src/index.ts
+++ b/packages/dtslint/src/index.ts
@@ -138,7 +138,7 @@ async function runTests(
     // so assert that we're really on DefinitelyTyped.
     assertPathIsInDefinitelyTyped(dirPath);
     assertPathIsNotBanned(dirPath);
-    assertPackageIsNotDeprecated(dirPath, await readFile("../notNeededPackages.json", "utf-8"));
+    assertPackageIsNotDeprecated(dirPath, await readFile(joinPaths(dirPath, "..", "..", "notNeededPackages.json"), "utf-8"));
   }
 
   const typesVersions = await mapDefinedAsync(await readdir(dirPath), async name => {


plumdog:
@AviVahl @andrewbranch Good chance I'm missing something, but this is my attempt. Does now run for me locally.

andrewbranch (changes requested):
This is going to be dependent upon whether the package being tested is in a versioned subfolder. We should probably just look up until we see the `types` directory and then go one more.

andrewbranch:
I think you and @sandersn were working on this at the same time: #405

plumdog:
Closing in favour of #405

PR #406, opened 2022-02-17
Fix incorrect relative path for package.json

maurizi:
This is causing DefinitelyTyped CI builds to fail.

Updating the relative path to match the one used on line 45 fixed this for me locally.

This seems to have been unintentionally been broken in https://github.com/microsoft/DefinitelyTyped-tools/pull/403


andrewbranch:
Thanks for the effort, but the path wasn’t wrong per se; the output structure of the package was.

PR #415, opened 2022-03-02
Add cssnano to the allowed dependencies

nicolas377:
cssnano now ships its own types, so this is setting the infrastructure up to deprecate @types/cssnano by adding `cssnano` to the allowed dependencies.


nicolas377:
For posterity's sake, this PR is for DefinitelyTyped/DefinitelyTyped#59155

PR #416, opened 2022-03-03
update(allowedPackageJsonDependencies): add nightwatch package

vaibhavsingh97:


sandersn:
Looking at the PR on Definitely Typed, I can't figure out which change added the dependency on `@nightwatch/react`. There aren't any explicit `import` statements that refer to it, are there?

vaibhavsingh97:
@sandersn I had added `package.json` file, and in there I had added `@nightwatch/react`
ref: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/59092/files#diff-7aba68c7f0b84bdfe2416b575e649596d798491f8e5f6b0fa26eae6b12ea201aR4

sandersn:
I saw the addition in package.json, but I couldn't figure out what code in the .ts files referred to `@nightwatch/react`. Another way of putting it: how does the compiler fail if `@nightwatch/react` is missing from package.json?

vaibhavsingh97:
@sandersn Thanks for raising the concern, so I went ahead and checked the code again, and found out that I really don't need `@nightwatch/react` as in the definitely typed project we only concerned about checking the correct types. This package is only needed when we are going to run `Nightwatch` tests.

I am closing this PR

PR #420, opened 2022-03-22
Add `chart.js` to allowed dependencies

DanGonite57:
For https://github.com/DefinitelyTyped/DefinitelyTyped/pull/59391

### DIFF
--- a/packages/definitions-parser/allowedPackageJsonDependencies.txt
+++ b/packages/definitions-parser/allowedPackageJsonDependencies.txt
@@ -347,6 +347,7 @@ canvas
 cash-dom
 cassandra-driver
 chalk
+chart.js
 chokidar
 ci-info
 cids


sandersn (approved):
When did chartjs-plugin-trendline switch from chart.js@2 to chart.js@3? 1.0 came out 5 months ago, so I guess that 0.22 used chart.js@2 and that `@types/chartjs-plugin-trendline` was incorrect in saying that it was for chartjs-plugin-trendline@1.0 until your PR.

DanGonite57:
Yes, my understanding is that chartjs-plugin-trendline began supporting chart.js@3 with the 1.0 version. The addition of `@types/chartjs-plugin-trendline` is very new (merged yesterday), I think it was written while accidentally referring to the bundled `@types/chart.js` (which is targeting chart.js@2.9) instead of chart.js@3's built-in type definitions.

PR #428, opened 2022-04-07
Allow tests to depend on npm packages

jablko:
Should we allow `devDependencies` in `types/*/package.json`? The `package.json` in the DT repo, not the one e.g. generated and published to npm.
- [hapi](https://www.npmjs.com/package/hapi) is a deprecated npm package: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/59709
- `@types/swagger-hapi` doesn't depend on `@types/hapi` but its tests do: https://github.com/DefinitelyTyped/DefinitelyTyped/runs/5856100773?check_suite_focus=true#step:6:425
  ```sh
  npm view @types/swagger-hapi dependencies
  { '@types/swagger-node-runner': '*' } # <- No @types/hapi
  ```
- Adding a "production" dependency on `@types/hapi` to `types/swagger-hapi/package.json` pacifies the DT CI, but would also publish the unnecessary dependency.
- [Adding `devDependencies`](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/59709/commits/c63ad3b2e35c7183bca3d95b8f4a90e42af34a11#diff-a724707ba5035978f69d885c80eb31c3c70782b50fd1798aa3453052cfc2886dR2-R4) instead pacifies it without publishing the dependency, because dtslint-runner installs `devDependencies` already, if they were [allowed](https://dev.azure.com/definitelytyped/DefinitelyTyped/_build/results?buildId=127397&view=logs&j=8635df7d-ecde-52a7-c7a8-84b998e35690&t=c264d665-0b79-59b7-7e04-9965cbe74860&l=479).
- &therefore; This PR adds `devDependencies` to the list of fields that dtslint allows in `package.json`.

Are `@types/swagger-hapi`'s tests wrong to depend on types that the types themselves do not? I had a look, and testing that the `hapi` and `swagger-hapi` types [are compatible](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/swagger-hapi/swagger-hapi-tests.ts#L2) seems legit?

### DIFF
--- a/packages/dtslint/src/checks.ts
+++ b/packages/dtslint/src/checks.ts
@@ -38,6 +38,7 @@ export async function checkPackageJson(dirPath: string, typesVersions: readonly 
     switch (key) {
       case "private":
       case "dependencies":
+      case "devDependencies":
       case "license":
       case "imports":
       case "exports":


sandersn:
Another solution is to bring back hapi types. If I remember correctly, its removal has caused more than one problem. I think that's the right solution here.

I also think it's strange for a test to test types that aren't required by the package itself. But I guess that can be a result of thorough testing like in this case.

jablko:
@sandersn How about other cases, e.g. `@types/hapi__vision` [depends on](https://www.npmjs.com/package/@types/hapi__vision?activeTab=dependencies) Handlebars.js, but only for [its tests](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/63dccaf6c4dc285433a9725fc21b847601c66350/types/hapi__vision/hapi__vision-tests.ts#L8) (which aren't published).

If dtslint allowed `devDependencies`, we could unpublish that dependency, I think.

jablko:
@sandersn No my understanding [from the docs](https://hapi.dev/module/vision/api/?v=6.1.0#handlebars) is `@hapi/vision` doesn't depend on Handlebars.js, it's just one of the engines that it works with. I double-checked I can install it without `handlebars`:
```sh
$ npm install @hapi/vision
$ npm ls --all | grep handlebars
# No output: handlebars isn't installed.

sandersn:
In that case I do think it's weird for the tests to specifically include it, and it sounds like a maintenance burden when handlebars updates but `@hapi/vision` doesn't. I'd rather see it test a mock or some other fake kind of object.

jakebailey:
This is now legal post-pnpm, though still follows the rules for what are and aren't allowed direct deps (the allowlist in the repo).

PR #432, opened 2022-04-11
Require maintainer approval to merge DT PRs vs. run CI

jablko:
Currently we need to land allowed dependency PRs, e.g. https://github.com/microsoft/DefinitelyTyped-tools/pull/431, before some DT PRs can progress through the review process (before the CI will pass). Sometimes the review process obsolesces those allowed dependency changes, if the DT PR is abandoned or changes are requested. It'd be nice if the allowed dependency changes could be deferred until the DT PR is actually merged?

dt-mergebot already requires maintainer approval to add dependencies -> `package.json`. What do you think about dropping `allowedPackageJsonDependencies.txt` altogether? That would involve maintainer intervention at the *end* of the DT review process (dt-mergebot and `package.json` dependencies), vs. at the beginning (definitions-parser and `allowedPackageJsonDependencies.txt`)?


MichaelDeBoey (approved):

sandersn:
The current approval process requires an approval from the the microsoft org, which is needed since it adds a dependency to a package published by Microsoft. An approval from the DefinitelyTyped org is not good enough.

PR #435, opened 2022-04-13
Default not-needed version

jablko:
Like https://github.com/microsoft/DefinitelyTyped-tools/pull/142, can we get the not-needed stub version from the npm registry, and leave it out of `notNeededPackages.json`?

What I did was:
- Extract the logic from `check-parse-results.ts` and `crawl-npm.ts` that gets the first version of the target package to contain built-in declarations.
- Use that logic to add a new, `minTypedVersion` field to our cached npm info.
- Make `asOfVersion` optional in `notNeededPackages.json`, and default to the `minTypedVersion` of its `libraryName`.
  That's the desired value: https://github.com/microsoft/types-publisher/pull/415/files#diff-8b7574dd77ee07295cf9fd5da671115eabc66bd7b0996d06e4208c97ff27df90R130
- Move `major`, `minor`, `desc`, and `id`, which depend on `version`, from `PackageBase` -> `TypingsData`.
- Add `version` parameters to `createNotNeededPackageJSON()`, `publishNotNeededPackage()`, and `deprecateNotNeededPackage()`, for the version of the incipient stub.

Already, the version of the incipient stub isn't necessarily `notNeededPackages.json`'s `asOfVersion`: It's the `skipBadPublishes()` version, which is the max of `asOfVersion` and the latest published types version, + 1: https://github.com/microsoft/types-publisher/pull/592/files#diff-ebe506ff86d233c1d2a4ff8cb9c196b74e578fbdef4f6ac5d6fb347202962a93R83

Instead of patching the `NotNeededPackage` instance, first with the default, then in `skipBadPublishes()`, I made `skipBadPublishes()` take and return a version. It takes `asOfVersion` (default: `minTypedVersion`) --- it returns the actual incipient stub version.

### DIFF
--- a/packages/definitions-parser/src/check-parse-results.ts
+++ b/packages/definitions-parser/src/check-parse-results.ts
@@ -1,19 +1,6 @@
 import { ParseDefinitionsOptions } from "./get-definitely-typed";
 import { TypingsData, AllPackages, formatTypingVersion } from "./packages";
-import {
-  assertDefined,
-  best,
-  mapDefined,
-  nAtATime,
-  FS,
-  logger,
-  writeLog,
-  Logger,
-  Semver,
-  UncachedNpmInfoClient,
-  NpmInfoRawVersions,
-  NpmInfoVersion
-} from "@definitelytyped/utils";
+import { nAtATime, FS, logger, writeLog, Logger, UncachedNpmInfoClient } from "@definitelytyped/utils";
 
 export async function checkParseResults(
   includeNpmChecks: false,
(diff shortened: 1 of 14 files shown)


MichaelDeBoey (approved):

sandersn (changes requested):
I don't like this for a couple of reasons.

1. I think that requiring a target version from the person who submits a deprecation PR is a good idea because it makes them double-check the now-typed version of the source package and compare it to the version on DT.
2. I want to minimise optional properties on DT. They make it more confusing to do the right thing by copying+modifying an existing entry.

PR #436, opened 2022-04-21
Use valid semvers internally

jablko:
Can we remove `semver.ts`, and instead use the standard `semver` npm package, which is similar?

`semver.ts` doesn't handle some valid cases, e.g. https://github.com/andrewbranch/definitely-not-slow/commit/d9d0828664b076801520c6c38c808296646a3ae7#diff-02cbc9970890cd33cf15c9ffb2f983aa63db1d290949662f06c134dad8fc9270R192 whereas switching to the standard package allows versions to be parsed consistently.


andrewbranch:
Yes please, this would be great. If you fix the build error I’ll do a quick check and get this in. Thanks!

jablko:
- [x] Thank you! I've fixed the build error.

andrewbranch:
@sandersn any objections?

andrewbranch (approved):

PR #441, opened 2022-04-25
Update links to ancestral repos

jablko:
Freshen some links so they point to the current monorepo:
- Links to lint rule docs: https://github.com/microsoft/DefinitelyTyped-tools/blob/068061ac920efb5380bc8f838b455385dfbb2a7a/packages/dtslint/src/util.ts#L13
- npm info, e.g. [@definitelytyped/perf](https://www.npmjs.com/package/@definitelytyped/perf), [types-registry](https://www.npmjs.com/package/types-registry).


andrewbranch (approved):
Thanks!

PR #442, opened 2022-04-25
Add expectedFailures for 4.7

sandersn:
There are a lot of packages that depend on old versions of knex. And a couple of dependents of apollo-link, which is deprecated.


sandersn:
@weswigham you should know about this, even though github won't show you in the reviewers list (?).

PR #445, opened 2022-04-30
Registry API no longer needs %2f-encoding

jablko:
I didn't have any luck tracking down when exactly this changed, but the npm registry hasn't needed %2f-encoding for some time, I think? e.g. https://registry.npmjs.org/@types/node

Can we remove the `fullNpmName`/`fullEscapedNpmName` distinction, in view of the similar but unrelated `unescapeName()`: https://github.com/microsoft/DefinitelyTyped-tools/blob/575fd10bc80d3fb50135a429855a3b0864983a95/packages/definitions-parser/src/packages.ts#L203-L204

### DIFF
--- a/packages/definitions-parser/src/packages.ts
+++ b/packages/definitions-parser/src/packages.ts
@@ -227,11 +227,6 @@ export abstract class PackageBase {
     return getFullNpmName(this.name);
   }
 
-  /** '@types%2ffoo' for a package 'foo'. */
-  get fullEscapedNpmName(): string {
-    return `@${scopeName}%2f${this.name}`;
-  }
-
   abstract readonly major: number;
   abstract readonly minor: number;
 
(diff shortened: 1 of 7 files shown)


jablko:
Just cleans up. It avoids mismatches between regular and %2f-encoded keys in the npm info cache, but I haven't actually found such a one.

jablko:
@andrewbranch Here are two existing cases that aren't %2f-encoded:
- https://github.com/microsoft/DefinitelyTyped-tools/blob/8665394ff115f155c42743131c9d3d324a9f5935/packages/publisher/src/calculate-versions.ts#L70
- https://github.com/microsoft/DefinitelyTyped-tools/blob/8665394ff115f155c42743131c9d3d324a9f5935/packages/publisher/src/calculate-versions.ts#L85

sandersn (approved):

PR #446, opened 2022-05-03
Update tar dependency

### LINKED ISSUE #347: Security vulnerability in @definitelytyped/utils dependency `tar`
`@definitelytyped/utils` requires `"tar": "^2.2.2"` which has an [Arbitrary File Creation/Overwrite](https://github.com/advisories/GHSA-5955-9wpr-37jh) vulnerability.

Please consider upgrading it order to fix this vulnerability.

image

michaldudak:
Updated the old `tar` dependency to the latest version and used its new API.

The `@types/tar` package seems to be incorrect (it shows the old API), so I decided to remove it completely. As the `tar` package is used in just one place, I think it's not a big deal to use the untyped library.

I haven't been able to definitely verify if setting directory permissions still works. 7Zip on Windows shows full permissions on the packed directory, but I haven't checked it on Linux yet.

Closes #347

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1809,13 +1809,6 @@
   resolved "https://registry.yarnpkg.com/@types/minimist/-/minimist-1.2.2.tgz#ee771e2ba4b3dc5b372935d549fd9617bf345b8c"
   integrity sha512-jhuKLIRrhvCPLqwPcx6INqmKeiA5EWrsCOPhrlFSrbrmU4ZMPjj5Ul/oLCMDO98XRUIwVm78xICz4EPCektzeQ==
 
-"@types/minipass@*":
-  version "2.2.0"
-  resolved "https://registry.yarnpkg.com/@types/minipass/-/minipass-2.2.0.tgz#51ad404e8eb1fa961f75ec61205796807b6f9651"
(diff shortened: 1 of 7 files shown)


yarn.lock
```diff
@@ -8049,7 +8018,7 @@ tar@^4.4.12:
     safe-buffer "^5.1.2"
     yallist "^3.0.3"
 
-tar@^6.0.2, tar@^6.1.0:
+tar@^6.0.2, tar@^6.1.0, tar@^6.1.11:
```
sandersn, line 8021: Why are the old versions of tar still in yarn.lock?

sandersn (changes requested):
Looks good although I have a question about yarn.lock

@sheetalkamat is on DT this week, so I want to let her know before merging this, just in case it causes problems.

yarn.lock
michaldudak: These are transitive dependencies (@lerna/pack-directory depends on ^6.1.0, node-gyp on ^6.0.2, etc.)

sandersn (approved):

sandersn:
Actually, tar failures won't make dts-critic fail CI, and it'll show up in the overnight build. I'll be notified tomorrow if there are problems.

PR #450, opened 2022-05-05
Move skipBadPublishes() -> calculate-versions

jablko:
Can we move skipBadPublishes() to the calculate-versions script, instead of calculating the incipient stub version from cached npm info twice, in generate-packages and publish-packages?

skipBadPublishes() calculates the version of the not-needed stub that we're about to publish. It [depends on calculate-versions](https://github.com/microsoft/types-publisher/pull/592/files#diff-ebe506ff86d233c1d2a4ff8cb9c196b74e578fbdef4f6ac5d6fb347202962a93R75) for cached npm-info. For regular packages, calculate-versions is where we actually perform that step (of figuring out the version we're about to publish).

What I did was:
- Move skipBadPublishes() to calculate-versions.ts. Combine it and isAlreadyDeprecated() into fetchIncipientStubVersion().
- Add that version to versions.json (the calculate-versions output), alongside the regular packages' versions-to-publish.
- Use it to replace skipBadPublishes() in generate-packages and publish-packages.


jablko:
I've rebased this on https://github.com/microsoft/DefinitelyTyped-tools/pull/451 and added a couple of tests for fetchIncipientStubVersion(), similar to the tests for fetchTypesPackageVersionInfo() I propose adding in https://github.com/microsoft/DefinitelyTyped-tools/pull/464.

sandersn:
With the new method of deprecation, I think the right thing is to remove skipBadPublishes entirely since it existed to skip the fallout from the publish/deprecate race condition on npm.

jablko:
@sandersn Are there any other ways for a `@types` package version to get ahead of its not-needed version? The npm-naming rule should ensure we don't publish a minor version that doesn't already exist on npm, but there are some types in the repo from before that rule was a thing, e.g. [`angular-ui-router`](https://github.com/microsoft/DefinitelyTyped-tools/pull/448#issue-1225752460).

I can imagine some other ways in theory it could happen (the JS package adds types in a patch version and our patch version is already greater, they add types in a new major/minor but we publish a corresponding `@types` version before adding it to `notNeededPackages.json`).

You're right about removing `skipBadPublishes()` from `publish-packages.ts` since the new method of deprecation, but I wonder if its worth keeping for other cases?

PR #453, opened 2022-05-05
Fetch npm metadata concurrently

jablko:
Can we speed up calculate-versions (when our npm cache is empty) using nAtATime(), like [the retag script does](https://github.com/microsoft/DefinitelyTyped-tools/pull/12/commits/c4f8bafc98bf0639cea7a7a17afc6b2ed3f334b4#diff-24dadcd1469cefcdb4abe26917a899910e4baaf576344a8f1a8fa99dd02f8cd5R49)?

What I did was replace `await mapDefinedAsync(...)` (which is serial) with `(await nAtATime(10, ...)).filter(value => value)`.

Before it took [+/- 30 minutes](https://github.com/jablko/DefinitelyTyped-tools/runs/6292258629?check_suite_focus=true#step:8:9518) (to fetch npm info for ~9,000 types packages + stubs). With nAtATime(10) it took [~3 minutes](https://github.com/jablko/DefinitelyTyped-tools/runs/6313245388?check_suite_focus=true#step:8:9915).

### DIFF
--- a/packages/publisher/src/calculate-versions.ts
+++ b/packages/publisher/src/calculate-versions.ts
@@ -2,14 +2,7 @@ import process from "process";
 import { defaultLocalOptions, defaultRemoteOptions } from "./lib/common";
 import { ChangedPackages, ChangedPackagesJson, ChangedTypingJson, versionsFilename } from "./lib/versions";
 import { getDefinitelyTyped, AllPackages, NotNeededPackage, writeDataFile } from "@definitelytyped/definitions-parser";
-import {
-  mapDefinedAsync,
-  logUncaughtErrors,
-  loggerWithErrors,
-  FS,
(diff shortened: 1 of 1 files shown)


jablko:
I've rebased this on https://github.com/microsoft/DefinitelyTyped-tools/pull/460 and removed nAtATime(), because pacote handles concurrency internally via [`maxSockets`](https://github.com/npm/npm-registry-fetch#-optsmaxsockets).

sandersn (approved):
This seems like a good idea, but I'd want to stage a careful manual test of it, since it's something that only happens naturally a couple of times a year.
I don't have time for that right now, so we should wait to merge this until next week.

jakebailey:
This was done back in #771.

PR #454, opened 2022-05-05
Add TS4.8 to supported versions

sandersn:
For use after typescript's main branch switches to 4.8 at the start of the 4.7 RC period.


sandersn:
@DanielRosenwasser this is ready for whenever `main` switches over to 4.8.

PR #456, opened 2022-05-11
Delete create-search-index.ts

jablko:
We stopped uploading `data/search-index-min.json`, I think: https://github.com/microsoft/DefinitelyTyped-tools/pull/51/files#diff-de7516b7f088e6c0b8ec69aee38be709e7d309719cb6f2c504280cd3c257c05dL21

Can we also stop building it?


karlhorky:
Oh it looks like this search index is being relied upon by Algolia's `npm-search` index:

https://github.com/algolia/npm-search/blob/d99c31fbe733953d39e498f9c7d982b2e09f3458/src/typescript/index.ts#L24-L40

Is there a recommendation for getting this types information from somewhere else? Eg.

1. Is a package written in TypeScript?
2. If not, does the package have a corresponding `@types/*` package published to Definitely Typed?

Seems like there was also [TypeSearch](https://www.typescriptlang.org/dt/search) (and the [TypeScout](https://typescout.vercel.app/) interface for it), but it seems like this is either unavailable or not containing all information that is necessary.

cc @Haroenv

Haroenv:
Thanks for letting me know, I wasn't aware this wasn't updated anymore (should have had better monitoring 🤔). I guess there's alternative options (maybe GitHub api?)

karlhorky:
Another idea: the npmjs.com website displays this correctly:

Screenshot 2022-12-06 at 11 06 09

More info here from @ethomson: https://twitter.com/ethomson/status/1353763656379543553

PR #463, opened 2022-05-22
gapi.client.* updates

Maxim-Mazurok:
Similar to https://github.com/microsoft/DefinitelyTyped-tools/pull/311 and other PRs by me, removing APIs that are removed from google discovery, and adding new ones


sandersn (approved):

Maxim-Mazurok:
> Turns out the 4 deleted lines are actually still in use by packages on DT. I'm going to restore them.

Right, I forgot that it's a two step process.. I'll remove them once my PR to DT to remove them is merged.

If we could add support for wildcards in this file, it'd really help me.

sandersn:
> Right, I forgot that it's a two step process

Me too. Sorry for the inconvenience.

> If we could add support for wildcards in this file, it'd really help me.

That makes sense, but the point of the file is to prevent people adding dependencies without telling us, which is defeated by a wildcard. Also the code right now is *really* simple (literally `new Set(text.split(/\n/))` followed by `if (set.has(name))`).

PR #464, opened 2022-05-22
Add fetchTypesPackageVersionInfo() tests

jablko:
I added a couple of tests for fetchTypesPackageVersionInfo(), to exercise https://github.com/microsoft/DefinitelyTyped-tools/pull/447.
- I based this PR on https://github.com/microsoft/DefinitelyTyped-tools/pull/451, to mock npm registry responses the same way as in that PR (using pacote).
- I removed the `canPublish` argument and instead return both the published and incipient versions. The caller can decide which they want.
- Added a comment with some notes.


sandersn:
I'm slammed right now, but I'll take a look at your PRs when I have my Definitely Typed rotation week.

In the meantime, the top of my wishlist is adding eslint to CI runs so that we can start migrating off of tslint, especially our custom rules. But that requires a lot of testing to make sure performance isn't worse and results are the same.

jablko:
@sandersn Understood. Thank you!

PR #470, opened 2022-05-29
add @khmyznikov/pwa-install package to allowedPackageJsonDependencies

dartess:
This package contains a web component. It also contains types for them. But it does't contain (and should not contain by design) types for using this component with react and preact. So, I plan to add extra types for him in @types.

### DIFF
--- a/packages/definitions-parser/allowedPackageJsonDependencies.txt
+++ b/packages/definitions-parser/allowedPackageJsonDependencies.txt
@@ -25,6 +25,7 @@
 @jest/test-result
 @jest/transform
 @jest/types
+@khmyznikov/pwa-install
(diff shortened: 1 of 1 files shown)


andrewbranch:
Link your DT PR when it's up and we'll merge this 👍

dartess:
@[andrewbranch](https://github.com/andrewbranch) Seems to be mutually exclusive paragraphs.

I can't finish DT PR because the tests don't pass, that's why i'm here. But I can open a PR with broken tests as WIP. It is acceptable?

andrewbranch:
Yes, the DT PR doesn't need to be green, I just want a link to it so we know this change is definitely going to get used.

PR #471, opened 2022-05-30
Add deprecated field to not-needed package.json

jablko:
Currently we publish and deprecate not-needed stubs in two steps, and frequently the first step succeeds and the second doesn't, e.g. [`@types/asn1js` versions 3.0.0 - 3.0.6](https://www.npmjs.com/package/@types/asn1js?activeTab=versions). By adding the deprecated field to the package.json, this PR will publish and deprecate in a single step, eliminating these "bad publishes".

### DIFF
--- a/packages/publisher/src/generate-packages.ts
+++ b/packages/publisher/src/generate-packages.ts
@@ -210,20 +210,19 @@ function dependencySemver(dependency: DependencyVersion): string {
   return dependency === "*" ? dependency : "^" + formatTypingVersion(dependency);
 }
 
-export function createNotNeededPackageJSON({ libraryName, license, fullNpmName, version }: NotNeededPackage): string {
+export function createNotNeededPackageJSON(pkg: NotNeededPackage): string {
   const out = {
-    name: fullNpmName,
-    version: String(version),
-    typings: null, // tslint:disable-line no-null-keyword
-    description: `Stub TypeScript definitions entry for ${libraryName}, which provides its own types definitions`,
+    name: pkg.fullNpmName,
+    version: String(pkg.version),
+    description: `Stub TypeScript definitions entry for ${pkg.libraryName}, which provides its own types definitions`,
     main: "",
     scripts: {},
-    author: "",
-    license,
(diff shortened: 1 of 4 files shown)


packages/publisher/src/generate-packages.ts
```diff
@@ -210,20 +210,19 @@ function dependencySemver(dependency: DependencyVersion): string {
   return dependency === "*" ? dependency : "^" + formatTypingVersion(dependency);
 }
 
-export function createNotNeededPackageJSON({ libraryName, license, fullNpmName, version }: NotNeededPackage): string {
+export function createNotNeededPackageJSON(pkg: NotNeededPackage): string {
   const out = {
-    name: fullNpmName,
-    version: String(version),
-    typings: null, // tslint:disable-line no-null-keyword
```
jablko, line 217: Not only are the `typings` and `author` fields empty, they don't exist in the regular types package package.json, so I took the liberty of tidying up.

sandersn (approved):
This looks really promising if I interpreted it correctly, because `client.deprecate` is unreliable.

I want to find a package to test it with before I merge though.

sandersn:
I got checkParseResults running again and found agilite. I'll ship this and deprecate it tomorrow morning to test it.

jablko:
> I searched for documentation on the deprecated field being used to deprecate and couldn't find anything. Do you have a reference?

@sandersn The npm registry docs is the closest thing, I think. They document the [`deprecated` field](https://github.com/npm/registry/blob/master/docs/responses/package-metadata.md#:~:text=deprecated%3A%20the%20deprecation%20warnings%20message%20of%20this%20version) and that the version object will [contain fields from the package.json](https://github.com/npm/registry/blob/master/docs/responses/package-metadata.md#:~:text=the%20full%20version%20object%20will%20also%20contain%20any%20other%20fields%20the%20package%20publisher%20chose%20to%20include%20in%20their%20package.json%20file%20for%20that%20version.), i.e. the `deprecated` field will be propagated from the pacakge.json to the npm registry metadata.

PR #472, opened 2022-05-30
Publish older versions without touching latest tag

jablko:
Use the `defaultTag` option when publishing older versions to avoid changing the `latest` tag and then having to undo and retag the latest version.

I switched to using libnpmpublish because I couldn't pass the `defaultTag` option to npm-registry-client's publish() method. libnpmpublish is what the npm cli's publish command now uses.

I based this on https://github.com/microsoft/DefinitelyTyped-tools/pull/471 because that saved mocking `client.deprecate()` for the `publishNotNeededPackage()` test.


packages/publisher/package.json
```diff
@@ -10,13 +10,16 @@
     "@definitelytyped/header-parser": "workspace:*",
     "@definitelytyped/retag": "workspace:*",
     "@definitelytyped/utils": "workspace:*",
+    "libnpmpublish": "^9.0.3",
```
sandersn, line 13: I thought this change would let us stop depending on npm-registry-publish, but it just adds the new dependency.

packages/utils/src/npm.ts
```diff
@@ -1,77 +1,38 @@
...
-import { createTgz } from "./io";
-
-export const npmRegistryHostName = "registry.npmjs.org";
-export const npmRegistry = `https://${npmRegistryHostName}/`;
 
 export const cacheDir = joinPaths(process.env.GITHUB_ACTIONS ? joinPaths(__dirname, "../../..") : os.tmpdir(), "cache");
 
-type NeedToFixNpmRegistryClientTypings = any;
-
 export class NpmPublishClient {
```
sandersn, line 9: this only tags and deprecates now. I'd rename to NpmTagger or something
and we seem to mostly create it on-demand, so maybe callers should switch to calling `tag`/`deprecate` directly.

packages/publisher/src/publish-packages.ts
```diff
@@ -46,12 +46,13 @@ export default async function publishPackages(
     log("=== Publishing packages ===");
   }
 
-  const client = await NpmPublishClient.create(await getSecret(Secret.NPM_TYPES_TOKEN), undefined);
+  const token = await getSecret(Secret.NPM_TYPES_TOKEN);
+  const client = new NpmPublishClient(token);
```
sandersn, line 50: splitting this statement seems unnecessary

sandersn (changes requested):

packages/utils/package.json
```diff
@@ -25,13 +25,11 @@
...
-    "tar": "^6.2.0",
     "tar-stream": "^3.1.6",
     "which": "^4.0.0"
   },
   "devDependencies": {
     "@types/charm": "^1.0.6",
-    "@types/tar": "^6.1.9",
     "@types/tar-stream": "^3.1.3",
     "@types/which": "^3.0.2",
     "@qiwi/npm-types": "^1.0.3"
```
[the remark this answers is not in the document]
jakebailey: This is one step, I have more changes I was building on this PR; probably I'll send a new one instead that includes this one.

packages/publisher/package.json
jakebailey: Yeah, it's half a change; this PR removed some of the registry dep but not all.

PR #477, opened 2022-06-07
Allow @aws-sdk/client-s3 dependency

Takaitra:
Allow DefinitelyTyped types to depend on @aws-sdk/client-s3. This is to unblock https://github.com/DefinitelyTyped/DefinitelyTyped/pull/60684


andrewbranch:
aws-sdk has individually published clients? How did I miss this? 😱 This is great news.

PR #483, opened 2022-06-16
Use pacote second attempt

jablko:
@andrewbranch Thanks for trying https://github.com/microsoft/DefinitelyTyped-tools/pull/451, I've now moved the cache into `os.tmpdir()`, to address https://github.com/microsoft/DefinitelyTyped-tools/pull/480#issue-1266598918. Would you be willing to give it another go, when you have a chance?

### DIFF
--- a/packages/utils/src/npm.ts
+++ b/packages/utils/src/npm.ts
@@ -1,157 +1,16 @@
-import assert = require("assert");
-import { ensureFile, pathExists, readJson, writeJson, readFile } from "fs-extra";
+import * as os from "os";
+import process from "process";
+import { readFile } from "fs-extra";
 import RegClient from "@qiwi/npm-registry-client";
 import { resolve as resolveUrl } from "url";
 import { joinPaths } from "./fs";
-import { loggerWithErrors, Logger } from "./logging";
-import { mapToRecord, recordToMap } from "./collections";
-import { Fetcher, createTgz } from "./io";
-import { sleep, identity } from "./miscellany";
+import { Logger } from "./logging";
+import { createTgz } from "./io";
 
 export const npmRegistryHostName = "registry.npmjs.org";
 export const npmRegistry = `https://${npmRegistryHostName}/`;
-export const npmApi = "api.npmjs.org";
 
-const defaultCacheDir = joinPaths(__dirname, "..", "cache");
-const cacheFileBasename = "npmInfo.json";
-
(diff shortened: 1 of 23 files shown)


packages/utils/src/npm.ts
```diff
@@ -1,157 +1,19 @@
...
-  const out: string[][] = [];
-  for (let i = 0; i < names.length; i += chunkSize) {
-    out.push(names.slice(i, i + chunkSize));
-  }
-  return out;
-}
+export const defaultCacheDir = joinPaths(
+  process.env.GITHUB_ACTIONS ? joinPaths(__dirname, "..") : os.tmpdir(),
+  "cache"
+);
```
jablko, line 16: Moved the cache into `os.tmpdir()`, to address https://github.com/microsoft/DefinitelyTyped-tools/pull/480#issue-1266598918.

andrewbranch: Verified that Node can write here on the Functions host over SSH, so I’m open to giving this another try. Using pacote seems like a worthwhile simplification if we can get it to work on our infrastructure. This week is busy for me though. I saw @sandersn merged a PR in this repo today so maybe he has some time?

sandersn: I'm on DT next week so I'll look then.

andrewbranch:
yolo

jablko:
@andrewbranch Thanks! :pray:

andrewbranch:
I believe it didn’t crash, but it also didn’t have anything to publish.

image

andrewbranch:
Actually, it looks like this is the old code, isn’t it? Because it looks like you removed the "Writing npm cache" log. The new version got published at 10:58 so I thought it would have used it here.

jablko:
That's right, looks like the version before this PR. Maybe the next run will use the new version?

jablko:
I take that as a sign it's using the new version, filling the cache.

andrewbranch:
It’s still going! 😄

jablko:
https://github.com/microsoft/DefinitelyTyped-tools/pull/453#issue-1227242953 speeds that up +/- 10 times.

andrewbranch:
Oops, I should have just merged that one too

andrewbranch:
Seemed to work... will watch the next run to make sure it successfully uses the new cache it built

andrewbranch:
Yep. I think this is good. Thanks! 🎉

jablko:
Thanks a lot!

Issue #422, opened 2022-03-29
Dependency detected but not installed


eps1lon:
Example: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/59519 (failing in https://github.com/DefinitelyTyped/DefinitelyTyped/runs/5726980310?check_suite_focus=true)

`styled-components` was correctly identified to depend on `react-native`. However, dependencies of `styled-components` weren't installed so tests are failing.

eps1lon:
Also blocking https://github.com/DefinitelyTyped/DefinitelyTyped/pull/59691

reichhartd:
Duplicate of https://github.com/microsoft/DefinitelyTyped-tools/issues/413

Issue #413, opened 2022-02-28
tests miss dependency in styled-components-react-native


sandersn:
`test-all` should find a dependency from styled-components-react-native to styled-components, but misses it. Here is the code:

```ts
// tslint:disable-next-line:no-single-declare-module
declare module "styled-components/native" {
    // ...
```

Probably the code that inspects `declare module` doesn't expect nested modules.

PR #484, opened 2022-06-17
Don't omit dependencies if only some deep modules are declared

jablko:
The problem that https://github.com/microsoft/DefinitelyTyped-tools/pull/384#issue-1107485477 addressed is that `getModuleInfo()` returns `dependencies` that are types package names and `declaredModules` that are full (possibly deep) module specifiers. Then `getTypingInfo()` tries to exclude the declared modules from the dependencies: https://github.com/microsoft/DefinitelyTyped-tools/blob/c037daa09fdfb854535a521a4a0015e52a491628/packages/types-publisher/src/lib/definition-parser.ts#L100-L102

https://github.com/microsoft/DefinitelyTyped-tools/pull/384#issue-1107485477 addressed the mismatch by calling `rootName()` on the declared modules, so we're comparing package names to package names. The problem with that is some packages declare *some* deep modules while legitimately depending on others, e.g. `meteor-dburles-collection-helpers` [imports `meteor/meteor`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e8d83de7f5a49bc894ebcf9e063a300a02e4a01f/types/meteor-dburles-collection-helpers/mongo.d.ts#L2) but only [declares `meteor/dburles:collection-helpers`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e8d83de7f5a49bc894ebcf9e063a300a02e4a01f/types/meteor-dburles-collection-helpers/collectionHelpers.d.ts#L2). Consequently `meteor` is excluded from its dependencies and the DT CI fails because dtslint-runner doesn't install Meteor's transitive external dependencies ([MongoDB](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e8d83de7f5a49bc894ebcf9e063a300a02e4a01f/types/meteor-dburles-collection-helpers/collectionHelpers.d.ts#L2)).

I think the solution is, instead of comparing names to names, we need to compare (possibly deep) specifiers to specifiers.

What I did was:
- Add a test, based on `meteor-dburles-collection-helpers`, that fails.
- Make `getModuleInfo()` return full specifiers in `dependencies`, same as in `declaredModules`.
- Faithfully move `rootName()` from `module-info.ts` -> `definition-parser.ts` and do the excluding of the current package from dependencies in `getTypingInfo()` vs. `getModuleInfo()`.
- Move two tests, `versionTypeRefThrows()` and `selfVersionTypeRefAllowed()`, from `module-info.test.ts` -> `definition-parser.test.ts`, because `rootName()` is now called from `getTypingInfo()` vs. `getModuleInfo()`.
- Update the remaining `getModuleInfo()` tests for the fact it now returns full specifiers, and doesn't exclude the current package.

Fixes https://github.com/microsoft/DefinitelyTyped-tools/issues/413#issue-1154633594
Fixes https://github.com/microsoft/DefinitelyTyped-tools/issues/422#issue-1185060727


.github/workflows/check-parse-results.yml
```diff
@@ -0,0 +1,28 @@
...
+      - run: yarn build
+      - uses: actions/checkout@v3
+        with:
+          repository: DefinitelyTyped/DefinitelyTyped
+          path: DefinitelyTyped
+          # So we can diff the most recent commit and its parent.
+          # https://github.com/microsoft/types-publisher/pull/290
+          fetch-depth: 2
+      - name: Check parse results
+        run: node --require source-map-support/register packages/dtslint-runner/ --path DefinitelyTyped/
```
sandersn, line 23: dtslint-runner isn't the same as check-parse-results in definitionParser, is it?

sandersn (changes requested):
I think this makes sense but I don't understand what check-parse-results.yml is supposed to do, even after looking at the failed CI run.

.github/workflows/check-parse-results.yml
jablko: @sandersn Yeah, these days check parse results is only called from dtslint-runner:
```sh
git grep --files-with-matches checkParseResults
packages/definitions-parser/src/check-parse-results.ts
packages/dtslint-runner/src/prepareAffectedPackages.ts
packages/dtslint-runner/src/prepareAllPackages.ts
```

That's probably alright?

The workflow is meant to confirm check parse results and the DT repo are in sync, to avoid DT CI outages. This PR would've cause `checkParseResults()` (and every DT PR except https://github.com/DefinitelyTyped/DefinitelyTyped/pull/60857#issue-1275806812) to fail if we hadn't fixed up the DT repo first. Before the DT half of this PR landed, the workflow showed the DT repo was out of sync: https://github.com/microsoft/DefinitelyTyped-tools/runs/6970784854?check_suite_focus=true

sandersn: Will PRs change the result of checkParseResults? I think it's only DT PRs that will, in which case the right place for the check to run is dtslint-runner, both the CI version and the overnight version.

jablko: @sandersn Well, changes to `getTypingInfo()`/`getModuleInfo()` change the dependency graph (potentially) which changes `checkParseResults()` and the DT repo states it accepts/rejects. That's the case with this PR.

Whether it happens that often, maybe not. But it's inexpensive to check, I think, and much nicer than finding out in the DT CI, after the `@definitelytyped` tools are deployed to npm and every DT PR fails?

Whereas the unit tests confirm the implementation matches assumptions, actually running check parse results confirms the assumptions match reality (at least the ones we can quickly check, and that potentially impact every PR). I think it's worth it?

jablko:
I noticed that check parse results was once a script like the other publisher steps, so you could run it independently from dtslint-runner:
- https://github.com/microsoft/DefinitelyTyped-tools/pull/6/commits/d23bd765263fbc90340c8ebe028ce32399f014ba#diff-cbadef7b90b30704b8749c037ef6f08e64f106af06a85ada118db2b38b3e49e6L27-L37
- https://github.com/microsoft/DefinitelyTyped-tools/pull/6/commits/d23bd765263fbc90340c8ebe028ce32399f014ba#diff-a182e89b4cb593f1dbade9fc181c81b168e03f8b25b0bc9cddc3d3b0cab83b46L44

I've now added a commit to this PR to restore that and make it an entry point again, and then run just check parse results from the workflow.

I also moved check parse results from the definitions-parser to the dtslint-runner package, because that's the only place it's used, currently?

jablko:
(I've now based the test on `styled-components-react-native` vs. `meteor-dburles-collection-helpers` since that's what the issue this PR fixes is about. It's the same test, otherwise.)

sandersn (approved):

PR #489, opened 2022-06-28
Port two rules from tslint to eslint

sandersn:
1. I'm still running a local test-all of this to find any bugs.
2. no-dead-reference is a whole-program text-based rule that basically ignores eslint structure -- like ExpectRule, except much, much simpler.
3. no-const-enum is about the simplest rule possible, and it relies on the TS AST.
4. I haven't checked existing typescript-eslint rules to see if any might apply. I (and others) should probably do that before porting any more.

Thanks to @JoshuaKGoldberg for pairing with me on this!


sandersn:
The local run quickly points out: before this ships, exclusions from tslint.json need to be ported before shipping this change.

packages/dtslint/test/index.test.ts
```diff
@@ -80,3 +83,44 @@ describe("dtslint", () => {
...
+describe("eslint", () => {
+  const ruleTester = new ESLintUtils.RuleTester({
+    parser: "@typescript-eslint/parser",
+  });
+  ruleTester.run("no-dead-reference", noDeadReference, {
+    invalid: [
+      {
+        code: `
+export class C { }
+/// <reference types="terms" />
```
JoshuaKGoldberg, line 95: May I suggest adding one or two more test cases around having multiple `///` reference comments?

JoshuaKGoldberg (commented):
🚀

PR #498, opened 2022-07-09
Use content hash vs. freshness in publish registry

jablko:
Same as the [calculate versions and retag scripts](https://github.com/microsoft/DefinitelyTyped-tools/blob/90eac4dbf8544e1c9a070984ba90f17b4856c4a3/packages/retag/src/index.ts#L122-L130), when gathering npm metadata, always revalidate if the content hashes differ (fresh or not) and never revalidate if they match (stale or not). This will use the cache for types that haven't changed since the last run, hitting the origin (npm registry) less often. If we have the content hash from the DT repo and the cached metadata, we might as well use them.


sandersn (changes requested):
Compared to the retag script, how many accesses to npm do you expect this will save? I didn't think that many were happening on an average run.

The fact that this code is in the retag script already is a good argument for copying it here -- but I'd like a good way to test that this is working. I guess deploying and watching a publish or two is one way, but is there some pattern of usage that would empty a lot of the cache and make it obvious that it is correctly deciding when to re-download from npm? I want to be confident of this change since it could cause subtle problems if it's wrong.

PR #512, opened 2022-08-09
Allow expected failures in dependents

forivall:
This change should allow https://github.com/DefinitelyTyped/DefinitelyTyped/pull/61001 and https://github.com/DefinitelyTyped/DefinitelyTyped/pull/61620 to pass CI

### DIFF
--- a/packages/dtslint-runner/src/main.ts
+++ b/packages/dtslint-runner/src/main.ts
@@ -56,7 +56,10 @@ export async function runDTSLint({
   }
 
   const allFailures: [string, string][] = [];
-  const expectedFailures = getExpectedFailures(onlyRunAffectedPackages);
+  let expectedFailures = getExpectedFailures();
+  if (onlyRunAffectedPackages) {
+    expectedFailures = new Set(dependents.filter((name) => expectedFailures.has(name)));
+  }
 
   const allPackages = [...packageNames, ...dependents];
   const testedPackages = shard ? allPackages.filter((_, i) => i % shard.count === shard.id - 1) : allPackages;
@@ -161,11 +164,7 @@ export async function runDTSLint({
   return allFailures.length;
 }
 
-function getExpectedFailures(onlyRunAffectedPackages: boolean) {
-  if (onlyRunAffectedPackages) {
-    return undefined;
-  }
-
+function getExpectedFailures() {
   return new Set(
     (readFileSync(joinPaths(__dirname, "../expectedFailures.txt"), "utf8") as string)
       .split("\n")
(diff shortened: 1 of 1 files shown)


andrewbranch:
Ah, you managed to catch both me and @sandersn on vacation. I believe @sandersn is back tomorrow-ish and would know more about this than I would. Also cc @jakebailey who’s on DT this week.

forivall:
Awesome, no worries. Yeah, I have no idea if this is the correct solution. I just want to get those lodash changes in, and so looking for whatever way to get it done, and this conceptually might do the job, so why not.

Thanks! I'll ping Jake on the DT repo on those issues.

packages/dtslint-runner/src/main.ts
```diff
@@ -161,16 +161,16 @@ export async function runDTSLint({
...
   if (onlyRunAffectedPackages) {
     return undefined;
   }
 
   return new Set(
     (readFileSync(joinPaths(__dirname, "../expectedFailures.txt"), "utf8") as string)
       .split("\n")
-      .filter(Boolean)
       .map((s) => s.trim())
+      .filter(onlyRunAffectedPackages ? (line) => line && dependents.includes(line) : Boolean)
```
jakebailey, line 173: This function returns early if `onlyRunAffectedPackages` is true, so I'm not sure that this is having the affect you intend in that case.

jakebailey: Yeah, this PR is effectively a noop (besides running this after `trim`), as `onlyRunAffectedPackages ` is guaranteed to be false by this point, and therefore `Boolean` will be chosen.

packages/dtslint-runner/src/main.ts
```diff
@@ -161,16 +161,16 @@ export async function runDTSLint({
   return allFailures.length;
 }
 
-function getExpectedFailures(onlyRunAffectedPackages: boolean) {
+function getExpectedFailures(onlyRunAffectedPackages: boolean, dependents: readonly string[]) {
   if (onlyRunAffectedPackages) {
     return undefined;
   }
```
forivall, line 167: These lines were supposed to be removed. Idk why I missed that in the commit, but I'll update this.

packages/dtslint-runner/src/main.ts
forivall: Ah, yeah, that early exit conditional was supposed to be removed.

sandersn (approved):

jakebailey (approved):
I was going to try and make a test for this, but the runner doesn't have any tests anyway. I'll merge and release it, then; I'm pretty sure this is right.

PR #514, opened 2022-08-09
Converted dt-header from TSLint to ESLint

JoshuaKGoldberg:
~Just opening this as a draft for reference. I still need to run this on DefinitelyTyped itself to double-check it works as expected...~ ✅ 

Continues the work from #489 by converting the `dt-header` rule from TSLint to ESLint.

I believe I ran this on DefinitelyTyped locally via `dtslint-runner`:

```shell
yarn build
node dist/index.js --path ~/repos/DefinitelyTyped --selection all --nProcesses 7
```


packages/dtslint/src/util.ts
```diff
@@ -1,9 +1,14 @@
+import { ESLintUtils } from "@typescript-eslint/utils";
 import assert = require("assert");
 import { pathExists, readFile } from "fs-extra";
 import { basename, dirname, join } from "path";
 import stripJsonComments = require("strip-json-comments");
 import * as ts from "typescript";
 
+export const createRule = ESLintUtils.RuleCreator(
+  (name) => `https://github.com/microsoft/DefinitelyTyped-tools/tree/master/packages/dtslint/src/rules/${name}.ts`
```
sandersn, line 9: do we seriously have prettier set to require parens around single parameters? I need to change that.

packages/dtslint/test/index.test.ts
```diff
@@ -70,93 +67,16 @@ describe("dtslint", () => {
...
-    const tests = readdirSync(testDir).filter((x) => x !== "index.test.ts");
+    const tests = readdirSync(testDir); // .filter((x) => x !== "index.test.ts");
     for (const testName of tests) {
       const testDirectory = join(testDir, testName);
       if (existsSync(join(testDirectory, "tslint.json"))) {
         testSingle(testDirectory);
-      } else {
+      } else if (statSync(testDirectory).isDirectory()) {
         for (const subTestName of readdirSync(testDirectory)) {
           testSingle(join(testDirectory, subTestName));
```
sandersn, line 77: I don't understand how `testSingle` is testing eslint files, since it delegates most of its work to tslint's consoleTestResultHandler and runTest. Are the eslint tests still getting run?

packages/dtslint/test/index.test.ts
```diff
@@ -70,93 +67,16 @@ describe("dtslint", () => {
     });
   });
   describe("rules", () => {
-    const tests = readdirSync(testDir).filter((x) => x !== "index.test.ts");
+    const tests = readdirSync(testDir); // .filter((x) => x !== "index.test.ts");
```
sandersn, line 70: remove instead of commenting

packages/dtslint/test/dt-header.test.ts
```diff
@@ -0,0 +1,168 @@
+import { ESLintUtils } from "@typescript-eslint/utils";
+
+import * as dtHeader from "../src/rules/dt-header";
+
+const ruleTester = new ESLintUtils.RuleTester({
+  parser: "@typescript-eslint/parser",
+});
+
+ruleTester.run("no-const-enum", dtHeader, {
```
sandersn, line 9:
```suggestion
ruleTester.run("dt-header", dtHeader, {
```

packages/dtslint/test/dt-header.test.ts
```diff
@@ -0,0 +1,168 @@
...
+          line: 1,
+          messageId: "parseError",
+        },
+      ],
+      filename: "types/abc/index.d.ts",
+    },
+    {
+      code: `// Type definitions for dt-header 1.0
+// Project: https://github.com/bobby-headers/dt-header
+// Definitions by: Jane Doe <https://github.org/janedoe>
```
sandersn, line 44: in the absence of test names, maybe this would help call out the problem

```suggestion
// Definitions by: Jane Doe <https://github.ORG/janedoe>
```

packages/dtslint/test/dt-header.test.ts
```diff
@@ -0,0 +1,168 @@
...
+        {
+          column: 2,
+          data: {
+            expected: "/\\/\\/ Type definitions for (non-npm package )?/",
+          },
+          line: 1,
+          messageId: "parseError",
+        },
+      ],
+      filename: "types/abc/index.d.ts",
```
sandersn, line 23: can you give a meaningful name instead of `abc`, since eslint doesn't let you name the test cases explicitly?

packages/dtslint/test/dt-header.test.ts
```diff
@@ -0,0 +1,168 @@
...
+`,
+      errors: [
+        {
+          column: 1,
+          endColumn: 27,
+          line: 3,
+          messageId: "definitionsBy",
+        },
+      ],
+      filename: "types/bad-url/index.d.ts",
```
sandersn, line 92:
```suggestion
      filename: "types/bad-username/index.d.ts",
```

packages/dtslint/test/dt-header.test.ts
```diff
@@ -0,0 +1,168 @@
...
+            expected: "foo MAJOR.MINOR (patch version not allowed)",
+          },
+          line: 1,
+          messageId: "parseError",
+        },
+      ],
+      filename: "types/foo/index.d.ts",
+    },
+    {
+      code: `// Type definitions for
```
sandersn, line 113: https://www.youtube.com/watch?v=mUj88_aPNkg

sandersn (changes requested):
Looks good, although I can't tell how the eslint tests are getting run right now. Just a few suggestions besides that.

Also, this needs to have an associated DT PR that converts the 1236 (!) exemptions to this rule and turns it on by default. That's going to happen often enough that we should maybe check in a tslint.json&rarr;.eslintrc script either here or in DT

packages/dtslint/test/index.test.ts
JoshuaKGoldberg: They are; Jest picks up any `test/**/*.test.ts` file in all packages based on the Jest config: https://github.com/microsoft/DefinitelyTyped-tools/blob/8b1430298dfffc2e235af5515f50094dfc512544/jest.config.js#L5

The CI output has a mention (https://github.com/microsoft/DefinitelyTyped-tools/runs/7754177707?check_suite_focus=true#step:7:17):

```plaintext
PASS packages/dtslint/test/dt-header.test.ts
```

packages/dtslint/src/util.ts
JoshuaKGoldberg: I'm just glad you're using Prettier in the first place 😄

packages/dtslint/test/index.test.ts
sandersn: cool, so this file can be deleted when the last of the tslint tests is removed!

sandersn (approved):

PR #518, opened 2022-08-12
Add 'type-fest' to allowed dependencies

Cattttttttt:
For https://github.com/DefinitelyTyped/DefinitelyTyped/pull/61681

'rollup-plugin-generate-package-json' requires 'read-pkg' and 'write-pkg', and these packages require 'type-fest'.


sandersn:
We rejected this dependency previously, I think because type-fest has many complex types besides PackageJson. See https://github.com/DefinitelyTyped/DefinitelyTyped/pull/59806#pullrequestreview-942584759

PR #523, opened 2022-08-18
Converted trim-file from TSLint to ESLint 

AlexMunoz:
Migration of another rule from tslint to eslint.

PR #521, opened 2022-08-18
Converted no-bad-reference from TSLint to ESLint

JoshuaKGoldberg:
Continues the work from #517, which continues the work from #514, which continues the work from #489.

Sibling PR to: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/62141.

PR #517, opened 2022-08-11
Converted export-just-namespace from TSLint to ESLint

JoshuaKGoldberg:
Continues the work from #514, which continues the work from #489.

Sibling PR to: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/62141.

PR #524, opened 2022-08-19
Remove dt-header from tslint config

jakebailey:
The rule is now in `.eslintrc.json` files, so I _think_ this is right and will get rid of these messages on DT:

```
Could not find implementations for the following rules specified in the configuration:
    dt-header
Try upgrading TSLint and/or ensuring that you have all necessary custom rules installed.
If TSLint was recently upgraded, you may have old rules configured which need to be cleaned up.
```

/cc @JoshuaKGoldberg

### DIFF
--- a/packages/dtslint/dt.json
+++ b/packages/dtslint/dt.json
@@ -1,7 +1,6 @@
 {
 	"extends": "./dtslint.json",
 	"rules": {
-		"dt-header": true,
 		"no-bad-reference": true,
 		"no-declare-current-package": true,
 		"no-self-import": true,


packages/dtslint/dt.json
```diff
@@ -1,7 +1,6 @@
 {
 	"extends": "./dtslint.json",
 	"rules": {
-		"dt-header": true,
```
JoshuaKGoldberg, line 4: Ah, yes - I'd missed this. Thanks!

JoshuaKGoldberg (approved):
Yes! 💯

JoshuaKGoldberg:
Thinking out loud: this should be done in #523, #521, and #517.

PR #531, opened 2022-09-01
[publisher] use consistent order for dependencies in the generated README

k-yle:
The order of dependencies in the README of `@types/...` packages changes depending on how the object is constructed in [this function](https://github.com/microsoft/DefinitelyTyped-tools/blob/master/packages/publisher/src/generate-packages.ts#L172-L194). 

This PR changes the order so that it is always consistent. This reduces the diff when a package is updated.


andrewbranch (approved):
Thanks!

PR #538, opened 2022-09-26
Converted prefer-declare-function from TSLint to ESLint

JoshuaKGoldberg:
Continues the work from #489.


JoshuaKGoldberg:
Bump @andrewbranch - is there any chance this & the other open TSLint->ESLint PRs be looked at soon, please?

sandersn:
@JoshuaKGoldberg I work on these during my DT week but I don't think anybody else is. When everybody gets back I'll ask others to take time as well.

packages/dtslint/src/rules/prefer-declare-function.ts
```diff
@@ -0,0 +1,34 @@
...
+      variableFunction: "Use a function declaration instead of a variable of function type.",
+    },
+    schema: [],
+    type: "problem",
+  },
+  name: "prefer-declare-function",
+  create(context) {
+    return {
+      // eslint-disable-next-line @typescript-eslint/naming-convention
+      "ExportNamedDeclaration > VariableDeclaration > VariableDeclarator"(node: TSESTree.VariableDeclarator) {
```
sandersn, line 22: the type doesn't need to be exported to be an error.

sandersn (changes requested):
The rule should apply to all variable declarations, not just exported ones.
I'll make the change since it's a simple one and this PR has been unreviewed for a long time.

packages/dtslint/src/rules/preferDeclareFunctionRule.ts
```diff
@@ -1,36 +0,0 @@
...
-    "Use a function declaration instead of a variable of function type."
-  );
-
-  apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
-    return this.applyWithFunction(sourceFile, walk);
-  }
-}
-
-function walk(ctx: Lint.WalkContext<void>): void {
-  eachModuleStatement(ctx.sourceFile, (statement) => {
```
sandersn, line 27: I missed this call to eachModuleStatement. It

1. Only runs on statements in declaration files.
2. Recursively runs on `declare module "foo"` blocks. This includes namespaces too, as far as I can tell.

The selector on line 22 is probably good enough for (2), but I'm not sure how to make (1) happen yet.

sandersn (approved):

sandersn:
I'll ship this tomorrow so I can create my fixup PR on DT right after.

PR #552, opened 2022-10-22
Add @hotwired/turbo as allowed dependency

kzkn:
For https://github.com/DefinitelyTyped/DefinitelyTyped/pull/62834


andrewbranch (approved):
It’s weird that @hotwired/turbo ships its own types but @hotwired/turbo-rails doesn’t. I thought maybe instead of putting them on DT someone should just ask the repo maintainers to ship them. But it looks like that was already done 2 years ago 🤦🏻‍♂️ https://github.com/hotwired/turbo-rails/issues/18

PR #562, opened 2022-11-07
Update to `@azure/functions@3`

sandersn:
Might be needed to update the Function runtime to version 4.


sandersn:
This should not ship until the Function runtime is updated from 3 to 4, and might not even be needed then. Probably a good idea though.

andrewbranch (approved):
Surprised there are no code changes necessary. Cool.

sandersn:
Merged. I'll publish this and try functions runtime@4. If that fails I'll revert and add the application insights upgrade I have sitting on my local machine.

PR #565, opened 2022-11-10
Add codeql scanning

sandersn:
Copied from the TS repo. I hope that the config file supports globs!

### DIFF
--- a/.github/codeql/codeql-configuration.yml
+++ b/.github/codeql/codeql-configuration.yml
@@ -1,0 +1,4 @@
+name : CodeQL Configuration
+
+paths:
+  - './packages/*/src' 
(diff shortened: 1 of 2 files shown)


sandersn:
So:
1. Overall working. There are two problems to investigate.
2. There is a problem finding the base branch. I don't know what this is.

sandersn:
The first check is not a problem.

1. The check is not smart enough to know that the division of the generated value exactly matches its range. It's only a problem when it doesn't -- modulus is the main offender here.
1. The random value is used in a shuffle to determine which files to test for performance. Even if the perf bot on DT weren't dead, there's no security problem in randomly choosing to test some files more often than others.

sandersn:
The second check is a real problem&mdash;we execute a string from the command line arguments&mdash;but it's in test code, so I'm not going to fix it.

PR #571, opened 2022-11-29
chore[definitions-parser]: whitelist `@types/react`

NivaldoFarias:
Refer to [New types definition for `next-pwa` - PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/63391) for an overview of why the `@types/react` package is needed and should be added.


NivaldoFarias:
Not needed anymore! [DT maintainer revised his decision](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/63391#discussion_r1036146397) to not include the triple-dash reference, which would mean that a `package.json` reference to `@types/react` would be needed. By using the triple-dash reference, no explicit addition to this file is needed.

PR #583, opened 2022-12-22
Add multiple `workbox` bundles to allowed packages

NivaldoFarias:
Resolving the issue mentioned at the [workbox-build PR](https://github.com/microsoft/DefinitelyTyped-tools/pull/581).

- [CI job error](https://github.com/DefinitelyTyped/DefinitelyTyped/commit/113bfcaa0694766600528aed13bd1f0b443abcbd/checks?check_suite_id=10019751963#step:6:14)


andrewbranch:
Are all of these _direct_ dependencies of something in DT? Transitive dependencies don’t need to be listed.

NivaldoFarias:
>

NivaldoFarias:
> Are all of these _direct_ dependencies of something in DT? Transitive dependencies don’t need to be listed.

That got me thinking, since the dependencies at DT `workbox-sw@4.3.2` points to other `workbox-_` DT packages _(most are at @4.3.2 aswell)_ that will be removed in favour of newer and bundled types by the original package. This may be a case (as mentioned in DT's [README](https://github.com/DefinitelyTyped/DefinitelyTyped#packagejson)) of adding the removed  `@types/workbox-_` packages to the dependencies list. 

Then, the right solution would be to add the @types/ packages instead of the original ones to the allowed packages `.txt` file. If this is correct and is approved, I will update both PRs.

packages/definitions-parser/allowedPackageJsonDependencies.txt
```diff
@@ -749,6 +749,19 @@
 @types/winston
 @types/wonder-commonlib
 @types/wonder-frp
+@types/workbox-background-sync
```
NivaldoFarias, line 752: In order to reference the type declarations used for the DT `customize-cra` and `workbox-sw` packages, all of these `@types/` packages need to be included in the allowed packages list. The dependencies inside the mentioned DT packages refer to older, to-be-removed DT `workbox-_` packages.

###### see [PR to remove old workbox-_ packages](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/63682)

sandersn (approved):

PR #587, opened 2023-01-04
Add `esbuild` to allowed packages

ciaran1344:


sandersn (changes requested):
Can you link to the Definitely Typed PR that adds a dependency on esbuild?

ciaran1344:
The module I was submitting types for now provides native typing, so this isn't needed

PR #614, opened 2023-02-24
Report results from running on-demand DT to github

gabritto:


packages/dtslint-runner/src/post-results.ts
```diff
@@ -0,0 +1,151 @@
...
+  const nightlyMap = new Map(nightly.map(error => [error.path, error]));
+  const branchMap = new Map(branch.map(error => [error.path, error]));
+
+  const nightlyOnly = [];
+  const bothChanged = [];
+
+  for (const [path, error] of nightlyMap) {
+    if (branchMap.has(path)) {
+      const branchError = branchMap.get(path)!;
+      if (branchError.message !== error.message) {
```
sandersn, line 100: How much noise does this suppress? I assume that it's better than `diff` that we're currently using.

packages/dtslint-runner/src/post-results.ts
```diff
@@ -0,0 +1,151 @@
...
+</details>
+` : "";
+  const bothChangedMessage =
+    bothChanged.length ? `
+<details>
+<summary>Errors that changed:</summary>
+${bothChanged.map(err => `Package: ${err.path}\nNightly error: ${err.nightly}\nBranch error: ${err.branch}`).join("\n")}
+</details>
+` : "";
+  const branchOnlyMessage =
```
sandersn, line 137: these are likely to be the actually valuable errors. Maybe this should be first, and not inside a `details` tag if there are only 1 - 3 errors?

sandersn (changes requested):
Some minor questions and comments. Looks good overall.

packages/dtslint-runner/src/post-results.ts
```diff
@@ -0,0 +1,151 @@
+import { Octokit } from "@octokit/rest";
+import { readFileSync } from "fs";
+import glob = require("glob");
+
+type Errors = { path: string, message: string }[];
+
+// Args: [auth token] [buildId] [status comment] [user to tag] [issue] [job status] [?nightly errors file] [?branch errors file]
```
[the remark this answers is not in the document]
gabritto: Yes

packages/dtslint-runner/src/post-results.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  try {
+    let newComment;
+    if (status === "fail") {
+      newComment = `Hey @${userToTag}, it looks like the DT test run failed. Please check the log for more details. ${checkLogsMessage}`;
+    }
+    else {
+      const nightlyErrors: Errors = [];
+      if (nightlyErrorsPath) {
+        const nightlyFiles = glob.sync(`**/*.json`, { cwd: nightlyErrorsPath, absolute: true });
```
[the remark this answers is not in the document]
gabritto: 1. They're named `nightlyFailuresN.json`, where `N` is the shard number.
2. The nightly and branch runs will be stored in different folders.

sandersn:
Update from collecting performance statistics from DT: two runs of the same commit showed a significant overall 2.3% speedup on. Another two runs showed a significant 20% speedup in ResolveTypeReference, which is a very short phase of compilation. I don't think the results are going to be reliable enough to use.

gabritto:
Some test (dtslint/dt-header) is failing, I have no idea why, but I'll investigate once I can get a successful pipeline run.

packages/dtslint/test/dt-header.test.ts
```diff
@@ -28,16 +28,37 @@ ruleTester.run("dt-header", dtHeader, {
       `,
       errors: [
         {
-          column: 2,
+          column: 1,
```
gabritto, line 31: This PR updates the version of `parsimmon`, a dependency of `header-parser`, from 1.13 to 1.18. Version 1.17 of `parsimmon` changed reporting of error line/column, and so this test needs to be updated to reflect that.

packages/dtslint/test/dt-header.test.ts
```diff
@@ -28,16 +28,37 @@ ruleTester.run("dt-header", dtHeader, {
...
             expected: "/\\/\\/ Type definitions for (non-npm package )?/",
           },
-          line: 1,
+          line: 2,
           messageId: "parseError",
         },
       ],
       filename: "types/only-comment/index.d.ts",
     },
+    {
```
gabritto, line 41: As per the above comment, I was wondering if `parsimmon` changed how they parse whitespace (maybe they were now ignoring it?), but it seems they didn't, and this test makes sure of that. So it seems the only change is in the position of the reported error.

packages/dtslint-runner/src/main.ts
```diff
@@ -63,7 +63,8 @@ export async function runDTSLint({
   const expectedFailures = getExpectedFailures(onlyRunAffectedPackages, dependents);
 
   const allPackages = [...packageNames, ...dependents];
-  const testedPackages = shard ? allPackages.filter((_, i) => i % shard.count === shard.id - 1) : allPackages;
+  let testedPackages = shard ? allPackages.filter((_, i) => i % shard.count === shard.id - 1) : allPackages;
+  testedPackages = testedPackages.slice(0, 4); // TODO: remove this line after testing, and make `testedPackages` a const.
```
sandersn, line 67: remember to remove this line

sandersn (approved):
Remember to revert the `slice(0,4)` code used for testing.

PR #626, opened 2023-03-01
Add @dcloudio/types to allowed dependencies

yurj26:
Required for:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/64519
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/64523

### DIFF
--- a/packages/definitions-parser/allowedPackageJsonDependencies.txt
+++ b/packages/definitions-parser/allowedPackageJsonDependencies.txt
@@ -8,6 +8,7 @@
 @babel/template
 @babel/traverse
 @babel/types
+@dcloudio/types
 @elastic/elasticsearch
 @electron/get
 @ember/test-helpers


andrewbranch:
Maybe @sandersn can weigh in, but this doesn’t seem necessary. The PRs you linked simply replace the entire contents of what was on DefinitelyTyped with a reference to @dcloudio/types. Users should simply use @dcloudio/types directly instead of going through an empty proxy package of @types/html5plus or @types/uni-app.

yurj26:
> Maybe @sandersn can weigh in, but this doesn’t seem necessary. The PRs you linked simply replace the entire contents of what was on DefinitelyTyped with a reference to @dcloudio/types. Users should simply use @dcloudio/types directly instead of going through an empty proxy package of @types/html5plus or @types/uni-app.

We have some old users who have been using @types/html5plus or @types/uni-app, we can't notify them to use @dcloudio/types, in order to allow users to continue to use, so we want to be consistent, thanks.
@andrewbranch @sandersn

PR #636, opened 2023-03-14
Drop tslint.json exact "extends" requirement outside of DefinitelyTyped

huntie:
The below validation rule, where any `tslint.json` file read by `dtslint` must contain `"extends": "@definitelytyped/dtslint/dtslint.json"`, is overly strict.

https://github.com/microsoft/DefinitelyTyped-tools/blob/9b08cd6c2bf48e8683527f4cb88d1bf29eb59e80/packages/dtslint/src/lint.ts#L209-L212

In this PR, I'm suggesting that this logic be reduced — validating that the JSON file is parsable only. (The strict check remains unchanged when `isDT` (within DefinitelyTyped repo) is `true`.)

**Motivation**

- In a monorepo setup, it's common to share config files by extending a file in a parent folder. This is allowed by `eslint`/`tslint`. However, when using `dtslint`, this convention breaks.

```js
{
  // The parent tslint.json *does* extend @definitelytyped/dtslint/dtslint.json
  "extends": "../../../tslint.json"
}
```

image

- A minimal config extending `@definitelytyped/dtslint/dtslint.json` is already [clearly documented in the README for new users](https://github.com/microsoft/DefinitelyTyped-tools/tree/master/packages/dtslint#typestslintjson) — I don't believe this restriction needs to exist within the tool.

**Test plan**

- When locally updated, command passes the `tslint.json` validation step (gets stuck on `tsconfig.json` step instead due to more direct validation of contents 😅).

### DIFF
--- a/packages/dtslint/src/lint.ts
+++ b/packages/dtslint/src/lint.ts
@@ -193,8 +193,6 @@ function testNoLintDisables(disabler: "tslint:disable" | "eslint-disable", text:
 export async function checkTslintJson(dirPath: string, dt: boolean): Promise<void> {
   const configPath = getConfigPath(dirPath);
   const shouldExtend = `@definitelytyped/dtslint/${dt ? "dt" : "dtslint"}.json`;
-  const validateExtends = (extend: string | string[]) =>
-    extend === shouldExtend || (!dt && Array.isArray(extend) && extend.some((val) => val === shouldExtend));
 
   if (!(await pathExists(configPath))) {
     if (dt) {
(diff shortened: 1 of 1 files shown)


huntie:
> seems like the validation is gone even for DT

True, actually that wasn't present beforehand — sorry for the incorrect description!

> Heads up: ...

Okay thanks! Maybe `eslint` is the best way forward then (I'd guessed that `dtslint` was required tooling for definitions).

sandersn:
No, it started that way and has migrated towards an internal DT-only tool. There are good typescript-eslint rules for checking types these days, which aren't tied to all our ramshackle infrastructure.

huntie:
Okay thanks. I think a big pull of sharing `dtslint` is for `dts-critic` behaviour.

We'll investigate looking into an equivalent `typescript-eslint` setup for both `facebook/react-native` and `facebook/metro`. cc @hoxyq @lunaleaps

PR #647, opened 2023-03-22
Allow --module node16

andrewbranch:
Related: https://github.com/DefinitelyTyped/dt-mergebot/pull/458

### DIFF
--- a/packages/dtslint/src/checks.ts
+++ b/packages/dtslint/src/checks.ts
@@ -56,16 +56,24 @@ export async function checkPackageJson(dirPath: string, typesVersions: readonly 
   }
 }
 
+/**
+ * numbers in `CompilerOptions` might be enum values mapped from strings
+ */
+export type CompilerOptionsRaw = {
+  [K in keyof CompilerOptions]?: CompilerOptions[K] extends number | undefined
+    ? string | number | undefined
+    : CompilerOptions[K];
+}
+
 export interface DefinitelyTypedInfo {
   /** "../" or "../../" or "../../../". This should use '/' even on windows. */
   readonly relativeBaseUrl: string;
 }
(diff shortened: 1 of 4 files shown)


packages/dtslint/src/checks.ts
```diff
@@ -56,16 +56,24 @@ export async function checkPackageJson(dirPath: string, typesVersions: readonly
   }
 }
 
+/**
+ * numbers in `CompilerOptions` might be enum values mapped from strings
```
sandersn, line 60: which specific properties have this? I assume it's "module".

packages/dtslint/src/checks.ts
```diff
@@ -56,16 +56,24 @@ export async function checkPackageJson(dirPath: string, typesVersions: readonly
...
+ */
+export type CompilerOptionsRaw = {
+  [K in keyof CompilerOptions]?: CompilerOptions[K] extends number | undefined
+    ? string | number | undefined
```
sandersn, line 64: do properties normally get literal union types instead of `string | number | undefined` ?

sandersn (approved):
Looks good:
1. A couple of questions about the new CompilerOptionsRaw type, although they're incidental.
2. I want to understand why all packages should eventually switch to node16, although shipping the ability for some to be node16 isn't worth waiting on that.

packages/dtslint/src/checks.ts
andrewbranch: `module` is the only one that matters here, but there are other enums that aren’t relevant to DT. Most numbers in `CompilerOptions` are actually enums represented by strings in the raw form. The only real number I can think of is `maxNodeModulesJsDepth`.

packages/dtslint/src/checks.ts
andrewbranch: Most of the ones that are getting transformed here are enum types, which are basically literal union types.

Issue #648, opened 2023-03-22
Port custom tslint rules to eslint


sandersn:
This is a tracking issue to co-ordinate work:

Done:

- no-outside-dependencies #650 
- no-import-default-of-export-equals
- no-self-import
- no-single-element-tuple-type
- no-relative-import-in-test
- no-redundant-jsdoc2 (though the new name should be no-redundant-jsdoc unless there's a built-in eslint rule already)

In-progress:

- no-single-declare-module

Not started:

- no-padding
- strict-export-declare-modifier
- npm-naming
- expect

Comment on this issue to let me know that you're working on a rule and I'll avoid working on it myself.

*Also* dtslint uses lots of default tslint rules. These all need eslint equivalents. I hope most (all?) of them are already written, but they all need to be tested and added to DT's .eslintrc.json.

JoshuaKGoldberg:
no-self-import looks quick, tackling it now...

JoshuaKGoldberg:
no-single-element-tuple-type too...

jakebailey:
If we're going to use `eslint-plugin-jsdoc`, we will definitely need to verify that it doesn't negatively impact performance. I removed this plugin from the TS repo for that reason; its parsing of JSDoc was extremely inefficient.

JoshuaKGoldberg:
> verify that it doesn't negatively impact performance

...indeed. In the `DefinitelyTyped` repo, after https://github.com/DefinitelyTyped/DefinitelyTyped/pull/65445:

<table>
<thead>
<tr>
<th>Variant</th>
<th>Timing</th>
</tr>
</thead>
<tbody>
<tr>
<td>Baseline</td>
<td><code>8.621 s ±  0.249 s</code></td>
</tr>
<tr>
<td><code>jsdoc</code></td>
<td><code>17.843 s ±  0.656 s</code></td>
</tr>
</tbody>
</table>

<details>
<summary>Getting performance measurements locally</summary>

After installing [`eslint-plugin-local-rules`](https://github.com/cletusw/eslint-plugin-local-rules):

```shell
npm i -D eslint-plugin-local-rules
```

I made the following two files:

<details>
<summary><code>.eslint-local-rules.cjs</code></summary>

```js
const dtHeader = require("../DefinitelyTyped-tools/packages/dtslint/dist/rules/dt-header");
const exportJustNamespace = require("../DefinitelyTyped-tools/packages/dtslint/dist/rules/export-just-namespace");
[...]

PR #653, opened 2023-03-24
Converted no-single-element-tuple-type from TSLint to ESLint

JoshuaKGoldberg:
Continues the work from #539. Ref: #648


sandersn (approved):
Looks good.

1. I want to get a clean run before merging.
2. This seems like another rule that might be generally useful.

sandersn:
OK, it's clean!

sandersn:
Well, I messed up the test run and I don't know why, but the new rule finds a lot more single element tuple types than the old did.
I'm running DT (with the rule **turned on this time**) and will add exemptions to the packages that fail.

PR #654, opened 2023-03-25
Converted no-relative-import-in-test from TSLint to ESLint

JoshuaKGoldberg:
~WIP. I can't figure out how to get the types to work in test.~ Strange, I'd been under the impression they were failing at some point. But it seems like they're working. ✅ 

Continues #648 by migrating `no-relative-import-in-test` from TSLint to ESLint.


sandersn:
I'm going to try to get this up-to-date and matched with a DT fixup today.

sandersn:
lodash is the only package with an exclusion for this rule.
However, with the exclusion removed, I can't get the rule to fail. I stuck some logging in and it looks like `ImportDeclaration` never gets hit. I think I'm unable to test locally with mismatching typescript versions. 

I'm going to merge this and ship it, but I need to figure out how to test locally reliably. It may be that dt-tools needs to update to the same version of typescript that DT uses.

PR #655, opened 2023-03-25
Converted no-single-declare-module from TSLint to ESLint

JoshuaKGoldberg:
Continues the work from #653. Ref: #648


packages/eslint-plugin/src/rules/no-single-declare-module.ts
```diff
@@ -0,0 +1,58 @@
+import { ESLintUtils } from "@typescript-eslint/utils";
+import * as ts from "typescript";
+
+import { createRule } from "../util";
+
+const rule = createRule({
```
jakebailey, line 6: Obligatory reminder to add this to the preset (tell me if you already know and I'll stop mentioning it in the future 😄)

sandersn: No, every single reminder is helpful.

sandersn (changes requested):
This is basically ready to go *except* that a local run is running out of memory much more often than before (it was 0 times before). I'm not sure what's going on; I need to test a version of this without the PR to make sure it isn't some other change.

I wonder whether it's possible that inspecting the sourcefile for module declarations and then converting that node back into the error-node somehow leaks the sourcefile node.

sandersn:
Happens on master too. I'll merge and ship this in the morning.

sandersn (approved):

sandersn:
Actually, uhhh, it's running out of memory on the initial

I checked the overnight run and on-demand test runs from yesterday, and no problems. I suspect this is to do with the use of TS ASTs though, now that there are two of them in the pipeline.

Or it could be the number of eslint rules all up, and adding two more has pushed some leak over the edge.

sandersn:
I'm going to try rewriting this rule to use the eslint tree and look generally eslinty, and see if that helps the problem.

PR #659, opened 2023-03-28
fix no-import-of-default-export

sandersn:
looks like it's `data: Record<string, any>` instead of `data: Record<string, string>`!

Edit: There are lots of failures overnight that don't happen on my machine. I'll fix that here too.


jakebailey (approved):

sandersn:
Nothing is on fire in the current local run, so I'm going to merge this and ship it to fix the overnight failures.

PR #663, opened 2023-04-10
PNPM workspaces

sandersn:
Current progress:
- builds
- tests pass
- I manually tested that dtslint succeeds on
   - 3box (some random package)
   - adeira__js (scoped package)
   - gulp/v3 (old-version package)
   - har-format (typesVersions package)
   - react
   - node
   react and node fail with errors now, which I need to look at.
- dtslint-runner parses all definitions on real DT.
- dtslint-runner succeeds in checking parse results.
- some packages fail, mostly with npm-naming and expect (I'm halfway through a local run)

TODO:

- #751
- Address a bunch of old TODOs
- Locally, force-publish every package and compare the output to a force-publish based on the present-day code.


packages/dtslint-runner/src/prepareAffectedPackages.ts
```diff
@@ -47,27 +43,15 @@ export async function prepareAffectedPackages({
...
-    // This doesn't work reliably on travis, and we're just installing for the types, so ignore.
-    const cmd = `npm install ${npmInstallFlags} --tag ts${pkg.minTypeScriptVersion}`;
-    console.log(`  ${cwd}: ${cmd}`);
-    const stdout = await execAndThrowErrors(cmd, cwd);
-    if (stdout) {
-      // Must specify what this is for since these run in parallel.
-      console.log(` from ${cwd}: ${stdout}`);
-    }
+  const cwd = definitelyTypedPath;
+  const cmd = `pnpm install`;
```
jakebailey, line 50: I suspect we shouldn't do this in the runner anymore, now that CI can do it, deal with caching, etc? Without a lockfile, this will incur the full cost of an install again, technically

sandersn: It should probably stick around to support the clone-DT-fresh-each-time scenario, but that should no longer be the default. I'm not even sure when it's useful, actually, because locally I always use my normal DT clone and don't care about `pnpm install` each time.

packages/header-parser/src/index.ts
```diff
@@ -85,6 +85,9 @@ export function validatePackageJson(packageName: string, packageJsonPath: string
         // "dependencies" / "license" checked by types-publisher,
         // TODO: asserts for other fields in types-publisher
         break;
+        // TODO: Only until Jake drops it
+      case "unmangledName":
```
jakebailey, line 89: This is gone, FWIW

packages/header-parser/src/index.ts
```diff
@@ -50,230 +56,297 @@
...
-        if (strict) {
-          return fail("patch version not allowed");
+  }
+  function validateVersion(): { major: number; minor: number } | { errors: string[] } {
+    const errors = [];
+    if (!packageJson.version || typeof packageJson.version !== "string") {
+      errors.push(
+        `${packageName}'s package.json should have \`"version"\` matching the version of the implementation package.`
+      );
+    } else if (!/\d+\.\d+\.\d+/.exec(packageJson.version)) {
```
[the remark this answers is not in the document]
sandersn: @jakebailey @andrewbranch do you think this is worth addressing? I can cut it down to 2 `\d+` occurrences by combining the next .99999 check with this one. And I could check that packageJson.version.length < 100. Or I could go see what `semver` does.

 But realistically, a catastrophically slow parse here means failed CI runs, a failed overnight run, and then some investigation to narrow down the offending package, followed by a revert.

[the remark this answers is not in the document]
jakebailey: I don't really know why we aren't just using `semver.parse` here...

packages/definitions-parser/src/mocks.ts
```diff
@@ -314,3 +303,24 @@ ${testNames.map((s) => "        " + JSON.stringify(s)).join(",\n")}
...
 }`;
 }
+
+function packageJson(packageName: string, version: string, dependencies: Record<string, string>) {
+  return `{
+    "private": true,
+    "name": "@types/${packageName}",
+    "version": "${version}.99999",
+    "projects": ["https://project"],
+    "contributors": [{
```
jakebailey, line 313: This should now be `owners`

packages/definitions-parser/src/packages.ts
```diff
@@ -181,65 +175,52 @@ function* flattenData(data: ReadonlyMap<string, TypingsVersions>): Iterable<Typi
...
-
 /** Prefer to use `AnyPackage` instead of this. */
 export abstract class PackageBase {
   static compare(a: PackageBase, b: PackageBase): number {
     return a.name.localeCompare(b.name);
   }
 
-  /** Note: for "foo__bar" this is still "foo__bar", not "@foo/bar". */
+  /** The package.json "name" field */
   abstract readonly name: string;
```
jakebailey, line 185: The meaning of this field is changing; is that intentional? Or maybe this already happened as a part of @andrewbranch's other PR?

jakebailey: Hm, reading `NotNeededPackage`, it doesn't have the `@types/` prefix? I am confused :D

packages/definitions-parser/src/packages.ts
```diff
@@ -507,59 +404,55 @@ export class TypingsVersions {
...
   }
   get minor(): number {
-    return this.data.libraryMinorVersion;
+    return this.data.header.libraryMinorVersion;
   }
 
   get minTypeScriptVersion(): TypeScriptVersion {
-    return TypeScriptVersion.isSupported(this.data.minTsVersion) ? this.data.minTsVersion : TypeScriptVersion.lowest;
+    return TypeScriptVersion.isSupported(this.data.header.typeScriptVersion)
+      ? this.data.header.typeScriptVersion
```
jakebailey, line 454: This is now `minimumTypeScriptVersion`

packages/definitions-parser/src/packages.ts
```diff
@@ -574,29 +467,29 @@ export class TypingsData extends PackageBase {
...
   }
-  get packageJsonDependencies(): readonly PackageJsonDependency[] {
-    return this.data.packageJsonDependencies;
+  get dependencies(): PackageJsonDependencies {
+    return this.data.dependencies ?? {};
+  }
+  get devDependencies(): PackageJsonDependencies {
+    return this.data.devDependencies ?? {};
+  }
+  *allPackageJsonDependencies(): Iterable<[string, string]> {
```
jakebailey, line 476: Iterators!! :tada:

packages/dtslint/src/checks.ts
```diff
@@ -69,76 +29,69 @@ export interface DefinitelyTypedInfo {
...
   if (!("lib" in options)) {
-    throw new Error('Must specify "lib", usually to `"lib": ["es6"]` or `"lib": ["es6", "dom"]`.');
+    errors.push('Must specify "lib", usually to `"lib": ["es6"]` or `"lib": ["es6", "dom"]`.');
   }
 
   if (!("module" in options)) {
-    throw new Error('Must specify "module" to `"module": "commonjs"` or `"module": "node16"`.');
-  }
-  if (
+    errors.push('Must specify "module" to `"module": "commonjs"` or `"module": "node16"`.');
```
jakebailey, line 84: Huh, since when could we do `node16` resolution?

packages/dtslint/src/lint.ts
```diff
@@ -1,7 +1,7 @@
 import { TypeScriptVersion } from "@definitelytyped/typescript-versions";
 import { typeScriptPath } from "@definitelytyped/utils";
 import assert = require("assert");
-import { pathExists } from "fs-extra";
+import { pathExistsSync } from "fs-extra";
```
jakebailey, line 4: Not for this PR, but modern Node's FS lib is rich enough that I suspect we could drop fs-extra.

packages/dtslint/src/lint.ts
```diff
@@ -112,11 +112,11 @@ function testDependencies(
     const typesFolder = dirname(path);
 
     return `
-A module look-up failed, this often occurs when you need to run \`npm install\` on a dependent module before you can lint.
+A module look-up failed, this often occurs when you need to run \`pnpm install\` on a dependent module before you can lint.
 
 Before you debug, first try running:
 
-   npm install --prefix ${typesFolder}
+   pnpm install --filter '{./types/${typesFolder}}...'
```
jakebailey, line 119: I'm actually somewhat suspicious of this rule anymore; what we should be telling people is to run `pnpm install -w --filter '...{./types/what-they-are-working-on}...'`, right? 

Or maybe it's unlikely people hit this at all anymore?

packages/dtslint/README.md
```diff
@@ -39,30 +39,76 @@ Refer to the [handbook](http://www.typescriptlang.org/docs/handbook/declaration-
...
+        "https://example.com/"
+    ],
+    "dependencies": {
+        "@types/DEPENDENCY-1": "*",
+        "@types/DEPENDENCY-2": "*"
+    },
+    "devDependencies": {
+        "@types/PACKAGE-NAME": "workspace:."
+    },
+    "contributors": [
```
jakebailey, line 80:
```suggestion
    "owners": [
```

packages/dtslint/README.md
```diff
@@ -108,15 +155,13 @@ f("one");
 ## Specify a TypeScript version
 
 Normally packages will be tested according to [DefinitelyType's support window](https://github.com/DefinitelyTyped/DefinitelyTyped#support-window).
-To use a newer version, specify it by including a comment like so:
+To restrict testing to new versions only, specify it in package.json:
 
 ```ts
-// Minimum TypeScript Version: 2.1
+"typeScriptVersion: 5.0"
```
jakebailey, line 161:
```suggestion
"minimumTypeScriptVersion: 5.0"
```

packages/dtslint-runner/src/check-parse-results.ts
```diff
@@ -1,217 +1,31 @@
...
+  const errors = [];
   for (const pkg of allPackages.allTypings()) {
     for (const dep of allPackages.allDependencyTypings(pkg)) {
-      if (dep.minTypeScriptVersion > pkg.minTypeScriptVersion) {
-        throw new Error(`${pkg.desc} depends on ${dep.desc} but has a lower required TypeScript version.`);
+      // check raw version because parsed version doesn't currently retain range information
+      const version = assertDefined(new Map(pkg.allPackageJsonDependencies()).get(dep.name));
+      if (semver.parse(version)) {
+        errors.push(
+          `${pkg.desc}'s dependency on ${dep.desc}@${version} must use "*" or a version range, not a specific version.`
```
jakebailey, line 20: Will this work for packages that were removed from DT but are being pinned back? I assume so given that weird express-brute-mongo thing works.

packages/dtslint-runner/src/prepareAllPackages.ts
```diff
@@ -1,27 +1,54 @@
...
+}
+const npmRetryCount = 5;
+export async function installAllDependencies(typesPath: string): Promise<void> {
+  console.log("Installing NPM dependencies...");
+  const cmd = `pnpm install ${pnpmInstallFlags}`;
+  console.log(`  ${typesPath}: ${cmd}`);
+  let lastError;
+  for (let i = 0; i < npmRetryCount; i++) {
+    try {
+      const stdout = await execAndThrowErrors(cmd, typesPath);
```
jakebailey, line 38: Is this execing at the types root, not the DT root? We probably want to run at the repo root, though maybe it's working because there's no package.json in that dir?

packages/header-parser/src/index.ts
```diff
@@ -1,42 +1,32 @@
...
+  readonly nonNpmDescription?: string;
+  readonly name: string;
   readonly libraryMajorVersion: number;
   readonly libraryMinorVersion: number;
   readonly typeScriptVersion: AllTypeScriptVersion;
   readonly projects: readonly string[];
   readonly contributors: readonly Author[];
 }
+// used in definitions-parser
+export type Author =
```
jakebailey, line 18: Annoyingly we're going to be calling this "owner" now. Oh well.

packages/publisher/src/util/util.ts
```diff
@@ -8,7 +7,5 @@ export function currentTimeStamp(): string {
 }
 
 export function outputDirectory(pkg: AnyPackage) {
-  return joinPaths(outputDirPath, pkg.desc);
+  return joinPaths(outputDirPath, pkg.typesDirectoryName + (pkg.isLatest ? "" : ` v${pkg.major}.${pkg.minor}`));
```
jakebailey, line 10: typesDirectoryName doesn't include the version?

packages/publisher/src/generate-packages.ts
```diff
@@ -123,83 +111,43 @@ async function outputFilePath(pkg: AnyPackage, filename: string): Promise<string
...
+export function createPackageJSON(typing: TypingsData, version: string): string {
   // Use the ordering of fields from https://docs.npmjs.com/files/package.json
   const out: {} = {
-    name: typing.fullNpmName,
+    name: typing.name,
     version,
     description: `TypeScript definitions for ${typing.libraryName}`,
     // keywords,
-    homepage: `https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/${typing.name}`,
+    homepage: `https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/${typing.typesDirectoryName}`,
```
jakebailey, line 121: In the previous comment we were adding a version to typesDirectoryName but now we're not; definitely should double check that the paths are what we expect.

packages/publisher/src/publish-registry.ts
```diff
@@ -144,7 +144,7 @@ async function installForValidate(log: Logger): Promise<void> {
     repository: {},
   });
 
-  const cmd = `npm install types-registry@next ${npmInstallFlags}`;
+  const cmd = `npm install types-registry@next ${pnpmInstallFlags}`;
```
jakebailey, line 147: This probably shouldn't get pnpm flags.

packages/retag/src/index.ts
```diff
@@ -139,8 +139,8 @@ export async function fetchTypesPackageVersionInfo(
   if (info.deprecated) {
     // https://github.com/DefinitelyTyped/DefinitelyTyped/pull/22306
     assert(
-      pkg.name === "angular-ui-router" || pkg.name === "ui-router-extras",
-      `Package ${pkg.name} has been deprecated, so we shouldn't have parsed it. Was it re-added?`
+      pkg.typesDirectoryName === "angular-ui-router" || pkg.typesDirectoryName === "ui-router-extras",
```
jakebailey, line 142: Not sure if `typesDirectoryName` includes a version or not, but maybe this should just be checking `=== "@types/angular-ui-router"`? Not sure what this check is doing at this point though.

packages/utils/src/assertions.ts
```diff
@@ -34,3 +34,20 @@ export function assertSorted<T>(
   }
   return a;
 }
+
+export function deepEquals(expected: unknown, actual: unknown): boolean {
```
jakebailey, line 38: Could this just be `assert.deepStrictEqual`?

packages/utils/src/io.ts
```diff
@@ -177,8 +177,7 @@ export async function isDirectory(path: string): Promise<boolean> {
   return (await stat(path)).isDirectory();
 }
 
-export const npmInstallFlags =
-  "--ignore-scripts --no-shrinkwrap --no-package-lock --no-bin-links --no-save --no-audit --no-fund --legacy-peer-deps";
+export const pnpmInstallFlags = "--ignore-scripts --no-save";
```
jakebailey, line 180: Unless we're using pnpm to install other stuff, we don't need to set ignore-scripts as that's already in npmrc. But, not harmful.

jakebailey: Yeah, in `typescript-installer.ts`, we're just using npm. It's possible that we just aren't doing enough installing anymore to warrant the constant?

jakebailey (commented):
Okay overall seems good, with some general notes of:

- Seems like there may be some inconsistency in the use of `.typesDirectoryName` and `.name`
- I think we should be able to drop parsimmon as a dep for the header parser now? May be worth running depcheck on the repo (or knip)

packages/dtslint/src/checks.ts
sandersn: uhh, I just remember porting this from the old code. This is a question for Andrew.

packages/dtslint-runner/src/check-parse-results.ts
sandersn: This rule is in here because I observed that `"*"` is the only thing used and I figured I want to keep it that way until there is a concrete reason to allow other things. But it was "empirically derived" meaning that I didn't stop to consider *what* other things might eventually be allowed. It might still be nothing, with overrides specified via pnpm overrides.

packages/definitions-parser/src/packages.ts
sandersn: No, it does. In NotNeededPackage's constructor:

```
this.name = name.startsWith(`@${scopeName}/`) ? name : `@${scopeName}/${name}`;
```

If you pass in a value with `@types/X`, you'll get that; otherwise `@types/` will be prepended.

sandersn: Oh, also, @andrewbranch's PR merged into this one. So it's part of his PR.

packages/definitions-parser/src/lib/utils.ts
```diff
@@ -25,3 +25,10 @@ export function withCache<T>(expiresInMs: number, getValue: () => Promise<T>): (
     return value!;
   };
 }
+
+export function slicePrefixes(s: string, prefix: string): string {
```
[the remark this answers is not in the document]
sandersn: Used in getTypingDataForSingleVersion and allDependencyTypings in definitions-parser.

packages/definitions-parser/src/lib/definition-parser.ts
```diff
@@ -174,84 +189,91 @@ export function parseVersionFromDirectoryName(
...
-): Promise<Omit<TypingsDataRaw, "libraryVersionDirectoryName">> {
-  const { remainingLs, typesVersions, hasPackageJson } = getTypesVersionsAndPackageJson(ls);
+): Promise<Omit<TypingsDataRaw, "libraryVersionDirectoryName"> | string[]> {
+  const errors = [];
+  const typesVersionAndPackageJson = getTypesVersionsAndPackageJson(ls);
+  if (Array.isArray(typesVersionAndPackageJson)) {
+    errors.push(...typesVersionAndPackageJson);
+  }
+  const { remainingLs, typesVersions, hasPackageJson } = Array.isArray(typesVersionAndPackageJson)
+    ? { remainingLs: [], typesVersions: [], hasPackageJson: false }
```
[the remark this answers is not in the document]
sandersn: I think this is gone now? Andrew must have removed it because grep can't find it.

packages/definitions-parser/src/git.ts
```diff
@@ -66,62 +56,74 @@ export async function gitDiff(log: Logger, definitelyTypedPath: string): Promise
...
-/** Returns all immediate subdirectories of the root directory that have changed. */
-export function gitChanges(diffs: GitDiff[]): PackageId[] {
-  const changedPackages = new Map<string, Map<string, DependencyVersion>>();
-
+/** Returns all immediate subdirectories of the root directory that have been deleted. */
+function gitDeletions(diffs: GitDiff[]): { errors: string[] } | { ok: PackageId[] } {
+  const changedPackages = new Map<string, PackageId>();
+  const errors = [];
   for (const diff of diffs) {
+    if (diff.status !== "D") continue;
```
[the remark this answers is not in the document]
sandersn: It will test things that might not need to be tested (but since they're complicated enough that git didn't understand, probably *should* be tested.) That's all that gitDeletions is used for now.

packages/definitions-parser/src/git.ts
```diff
@@ -19,9 +19,9 @@ export interface GitDiff {
 }
 
 /*
-We have to be careful about how we get the diff because travis uses a shallow clone.
+We have to be careful about how we get the diff because Actions uses a shallow clone.
```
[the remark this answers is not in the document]
sandersn: It's referring to how Actions (used to??) clone.

packages/definitions-parser/src/get-affected-packages.ts
```diff
@@ -1,109 +1,82 @@
...
-    add(item);
-  }
-
-  while (workList.length) {
-    const item = workList.pop()!;
-    for (const newItem of getRelatedItems(item)) {
-      add(newItem);
-    }
+  const changedPackageNames = await execAndThrowErrors(
+    `pnpm ls -r --depth -1 --parseable --filter '[${sourceRemote}/${sourceBranch}]'`,
```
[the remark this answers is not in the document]
sandersn: I should add a TODO to see whether we *really* need to distinguish edited packages from non-edited ones; we do have `...[remote/branch]` above in `const filters = [--filter '...`, and then we re-run `ls` just for the edited packages.

packages/definitions-parser/src/get-affected-packages.ts
```diff
@@ -1,109 +1,82 @@
...
+  const allDependents = [];
+  const filters = [`--filter '...[${sourceRemote}/${sourceBranch}]'`];
+  for (const d of deletions) {
+    for (const dep of allPackages.allTypings()) {
+      for (const [name, version] of dep.allPackageJsonDependencies()) {
+        if (
+          "@types/" + d.typesDirectoryName === name &&
+          (d.version === "*" || satisfies(formatTypingVersion(d.version), version))
+        ) {
+          filters.push(`--filter '...{./types/${dep.name}}'`);
```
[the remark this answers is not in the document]
sandersn: I think it will: dep.name should be, eg, react which I *think* will include react/v15 if v15 depends on the deleted package.

And if you delete an old version, then on the previous line with `satisfies`, both deletions and dep.allPackageJsonDependencies will correctly produce whatever/v2 for `d.version` and `version`, respectively.

packages/utils/src/assertions.ts
jakebailey: Ah, duh, that doesn't do a boolean.

packages/definitions-parser/src/lib/definition-parser.ts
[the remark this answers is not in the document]
jakebailey: Yep, he removed it.

packages/definitions-parser/src/lib/utils.ts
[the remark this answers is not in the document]
jakebailey: Oops, guess find-all-refs didn't work.

packages/definitions-parser/src/get-affected-packages.ts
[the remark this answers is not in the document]
jakebailey: Right, but I'm not sure what the `[...]` syntax does in pnpm if you don't actually do `...` to ask it to diff. May be no different than not passing a filter at all.

packages/dtslint-runner/src/check-parse-results.ts
jakebailey: Well, CI should tell us if we do things wrong at least.

packages/header-parser/src/index.ts
jakebailey: Yeah however you want, I am not picky. Don't know who all is using this code but I'm sure this change is breaking in every conceivable way to rename how you feel.

packages/publisher/src/generate-packages.ts
jakebailey: I guess that's true, yeah; obviously one can move v18 to a v18 dir and the paths shouldn't break or differ.

packages/definitions-parser/src/get-affected-packages.ts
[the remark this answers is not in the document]
jakebailey: But, I thought after Andrew's change, `dep.name` is now `@types/react`? I would expect this to be `dep.typesDirectoryName` as that's the path being used here.

packages/definitions-parser/src/packages.ts
jakebailey: Right, it's definitely the name with `@types/` now

jakebailey:
I am not sure where, but some part of the code's linting is still suggesting putting stuff in the header:

```
Error in wicg-file-system-access
Error: /home/runner/work/DefinitelyTyped/DefinitelyTyped/DefinitelyTyped/types/wicg-file-system-access/index.d.ts:1:1
ERROR: 1:1  npm-naming  Declaration file must have a matching npm package.
To resolve this error, either:
1. Change the name to match an npm package.
2. Add a Definitely Typed header with the first line


// Type definitions for non-npm package wicg-file-system-access-browser

Add -browser to the end of your name to make sure it doesn't conflict with existing npm packages.
If you won't fix this error now or you think this error is wrong,
you can disable this check by adding the following options to your project's tslint.json file under "rules":

    "npm-naming": false
 See: https://github.com/microsoft/DefinitelyTyped-tools/blob/master/packages/dtslint/docs/npm-naming.md

    at testTypesVersion (/home/runner/work/DefinitelyTyped/DefinitelyTyped/DefinitelyTyped-tools/packages/dtslint/src/index.ts:218:11)
    at runTests (/home/runner/work/DefinitelyTyped/DefinitelyTyped/DefinitelyTyped-tools/packages/dtslint/src/index.ts:186:7)

```

jakebailey:
yeah dts-critic's `checkNpm` still says this.

packages/definitions-parser/src/get-affected-packages.ts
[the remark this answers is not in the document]
andrewbranch: Or just `` `...${dep.name}` ``, right?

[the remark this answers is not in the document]
jakebailey: Probably that, since this doesn't seem to include versioning, may as well install the whole package even if we have more than one. Likely safer.

packages/definitions-parser/src/get-affected-packages.ts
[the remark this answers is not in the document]
sandersn: I edited react and manually tested 
`--filter '...[jakebailey/pnpm-workspaces-working]'`
`--filter '[jakebailey/pnpm-workspaces-working]'`
`--filter '[jakebailey/pnpm-workspaces-working]...'`
and they all gave the results I expected (react + dependents, react only, react + dependencies)

packages/definitions-parser/src/git.ts
```diff
@@ -66,62 +56,74 @@ export async function gitDiff(log: Logger, definitelyTypedPath: string): Promise
...
+  const errors = [];
   const diffs = await gitDiff(consoleLogger.info, definitelyTypedPath);
   if (diffs.find((d) => d.file === "notNeededPackages.json")) {
-    for (const deleted of getNotNeededPackages(allPackages, diffs)) {
-      checkNotNeededPackage(deleted);
+    const deleteds = getNotNeededPackages(allPackages, diffs);
+    if ("errors" in deleteds) errors.push(...deleteds.errors);
+    else
+      for (const deleted of deleteds.ok as NotNeededPackage[]) {
+        errors.push(...(await checkNotNeededPackage(deleted)));
```
andrewbranch, line 94: Not proposing to change it now, but it’s really weird that this `get*` function is _checking_ something as a side effect. It’s also weird that a function that’s supposed to “get affected packages” takes a parameter for whether you want... the affected packages... or all of them 🤔

andrewbranch (approved):
I've looked pretty thoroughly at

- definitions-parser
- header-parser

and personally tested publisher pretty thoroughly. I haven't looked closely at dtslint, dtslint-runner, or eslint-plugin, but you've been testing those on DT as we go. So, I think I feel pretty good.

packages/dts-critic/index.ts
```diff
@@ -233,34 +231,32 @@ Try adding -browser to the end of the name to get
...
       message: `Declaration file must have a matching npm package.
 To resolve this error, either:
 1. Change the name to match an npm package.
-2. Add a Definitely Typed header with the first line
-
-
-// Type definitions for non-npm package ${name}-browser
-
-Add -browser to the end of your name to make sure it doesn't conflict with existing npm packages.`,
+2. Add \`"nonNpm": true\` to the package.json to indicate that this is not an npm package.
```
jakebailey, line 241: Should we still suggest the `-browser` suffix as an example?

jakebailey (approved):
Seems good to me now, with the one comment about whether or not we should suggest a non-conflicting name in the new lint message.

jakebailey (approved):
Seems good to me now, with the one comment about whether or not we should suggest a non-conflicting name in the new lint message.

jakebailey:
Also, are we requiring `nonNpmDescription` when `nonNpm` is specified? I think before you _had_ to because it'd use the text from the header, and we probably still want that?

packages/definitions-parser/src/git.ts
sandersn: I also just noticed that `selection` can be a regex, and that there's code to run only packages that match it.

sandersn:
nonNpmDescription is required when nonNpm is specified, and not allowed when it is not.

packages/dts-critic/index.ts
jakebailey: Yeah, we can do this later. `-browser` leads to people tacking `-browser` to their name to silence the lint even if it doesn't make any sense... Saw that once already last week.

packages/definitions-parser/src/git.ts
sandersn: Moments later: This code is **only** called with "affected"; "all" and regexes are unused. I deleted the dead code.

packages/dts-critic/index.ts
andrewbranch: Yeah, I felt that advice was very outdated at best

PR #665, opened 2023-04-19
Add wordpress components to allowed dependencies

noahtallen:
`@wordpress/components` now ships types (see https://www.npmjs.com/package/@wordpress/components)

Since I'd like to remove the DT types here (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/65213), and since we now need to use npm to import this package for some other `@types/wordpress` packages, it should be added to this lsit

### DIFF
--- a/packages/definitions-parser/allowedPackageJsonDependencies.txt
+++ b/packages/definitions-parser/allowedPackageJsonDependencies.txt
@@ -568,6 +568,7 @@
 @wordpress/core-data
 @wordpress/data
 @wordpress/element
+@wordpress/components
 abort-controller
 acorn
 actions-on-google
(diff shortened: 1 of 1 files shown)


packages/definitions-parser/allowedPackageJsonDependencies.txt
```diff
@@ -568,6 +568,7 @@
 @wordpress/core-data
 @wordpress/data
 @wordpress/element
+@wordpress/components
```
jakebailey, line 571: This should be alphabetical.

noahtallen: done

jakebailey (approved):

jakebailey (approved):

PR #667, opened 2023-04-26
Add googleapis to allowed dependencies

KonradPerlicki:
Required for: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/65283

### DIFF
--- a/packages/definitions-parser/allowedPackageJsonDependencies.txt
+++ b/packages/definitions-parser/allowedPackageJsonDependencies.txt
@@ -667,6 +667,7 @@ gl-matrix
 glaze
 glob
 globby
+googleapis
 google-auth-library
 google-gax
 got


jakebailey (changes requested):
Er, aren't all Google APIs already covered by the gapi packages?

KonradPerlicki:
You're right, I have replaced the package, thanks and I'm closing PR

jakebailey:
On second look, I don't know that the gapi packages are the right ones because of their global behavior, actually...

jakebailey:
But maybe googleapis does that too. Not really sure what the difference is between them.

KonradPerlicki:
The original package uses googleapis so It was my first thought to use it as well

jakebailey:
Ah, then that settles that. 127MB package, though, wow!

KonradPerlicki:
So, will you accept that and should I revert my last changes from [previous](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/65283) or what now?

jakebailey:
Yes, given the npm package depends on `googleapis`.

jakebailey (approved):

PR #673, opened 2023-05-10
Remove applicationinsights

sandersn:
Based on an analysis of azure usage with @RyanCavanaugh.

This hasn't worked for a long time, because there's a bug in the publish notification code that runs before the call to application insights. I'm not sure whether to delete *that* or to fix it. Publication has been reliable (and predictable!--since the Functions changeover) for a long time so neither the notification nor the stats are as necessary as they once were.

### DIFF
--- a/packages/publisher/package.json
+++ b/packages/publisher/package.json
@@ -15,7 +15,6 @@
     "@definitelytyped/typescript-versions": "^0.0.159",
     "@definitelytyped/utils": "^0.0.159",
     "@octokit/rest": "^16.1.0",
-    "applicationinsights": "^2",
     "fs-extra": "^9.1.0",
     "fstream": "^1.0.12",
     "hh-mm-ss": "^1.2.0",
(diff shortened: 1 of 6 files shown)


jakebailey (approved):

sandersn:
Note to self: after I merge and ship this, I need to remove APPINSIGHTS_INSTRUMENTATIONKEY from Azure as well. (it silently fails to log without it, and we're already not logging successfully, so actually I guess it's OK to delete it whenever)

andrewbranch:
This is how we see logs when the publisher crashes. This was definitely working and useful fairly recently.

jakebailey:
Oh :(

andrewbranch:
To clarify, I don’t think we were using the events, but I believe the setup/start calls are necessary in order to see stdout/stderr in the Azure portal.

sandersn:
All right, I'll revert it tomorrow and then remove only the event logging.

PR #678, opened 2023-05-19
Fix RC installation

### LINKED ISSUE #677: During RC period, dtslint fails on packages with typesVersions of the version previous to the RC.
For example, when (1) react has a ts5.0 subdirectory (2) TS 5.1 is in RC, the versions tested are normally [4.3,4.4,4.5,4.6,4.7,4.8,4.9,5.0,5.2]. 5.1 is skipped. That's because, due to an implementation quirk, dtslint only tests 1 pre-release version: `typescript@next`, and both 5.1 and 5.2 are pre-release (5.1 is `typescript@rc`; 5.2 is `typescript@next`). 

Normally dtslint starts by testing two versions: lowest and highest. In this example, that's [4.3, 5.2]. If one of those fails, other versions will be tested to narrow down the first version with an error.

However, for packages with a typesVersion entry like ts5.0, dtslint tests versions up to ts*X.Y* on the tsX.Y files: [4.3, 5.0]. Then its tests newer versions on the toplevel files: [5.1, 5.2].

The problem when testing [4.3, 5.0, 5.1, 5.2] is that 5.1 is not installed. And it's not trivial to install either; it's only tagged as `typescript@rc`, but most of the time `typescript@rc` is irrelevant. You can distinguish the RC condition using `@definitelytyped/typescript-versions` -- when `supported` has two more entries than `shipped`. In that case dtslint could install `typescript@rc` as `supported.at(-2)`

sandersn:
Fixes #677
See that bug for discussion; it's somewhat complicated, but the fix is: during RC you have to install the RC as well, just in case.

### DIFF
--- a/packages/utils/src/typescript-installer.ts
+++ b/packages/utils/src/typescript-installer.ts
@@ -14,6 +14,7 @@ export async function installAllTypeScriptVersions() {
   for (const v of TypeScriptVersion.shipped) {
     await install(v);
   }
+  // `shipped + [rc, next] == supported` during the RC period. During that time, typescript@rc needs to be installed too.
   if (TypeScriptVersion.shipped.length + 2 === TypeScriptVersion.supported.length) {
     await install("rc");
   }
--- a/packages/utils/src/typescript-installer.ts
+++ b/packages/utils/src/typescript-installer.ts
@@ -14,6 +14,9 @@ export async function installAllTypeScriptVersions() {
   for (const v of TypeScriptVersion.shipped) {
     await install(v);
   }
+  if (TypeScriptVersion.shipped.length + 2 === TypeScriptVersion.supported.length) {
+    await install("rc");
+  }
   await installTypeScriptNext();
 }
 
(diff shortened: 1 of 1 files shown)


packages/utils/src/typescript-installer.ts
```diff
@@ -14,22 +14,32 @@ export async function installAllTypeScriptVersions() {
...
+  }
   await installTypeScriptNext();
 }
 
 export async function installTypeScriptNext() {
   await install("next");
 }
 
-async function install(version: TsVersion | "next"): Promise<void> {
+async function install(version: TsVersion | "next" | "rc"): Promise<void> {
```
jakebailey, line 27: One problem I see with installing `rc` without checking is that for 5.1, we added it to the array _before_ the rc was available, so if this runs, it'll get the previous rc from an old version and potentially make people think their code is working. Is that accurate?

sandersn: The check on line 17 means that it only happens during the RC period. I'll put a better comment.

PR #680, opened 2023-05-21
Converted strict-export-declare-modifiers from TSLint to ESLint, ish

JoshuaKGoldberg:
Continues #648 by migrating `strict-export-declare-modifiers` from TSLint to ESLint...

...mostly. This removes the dependency on `tslint` but still operates on the TS tree instead of TSESTree tree. I ran out of energy overhauling these more TypeScript-intertwined rules... is this ok as an intermediary step?


packages/eslint-plugin/src/rules/strict-export-declare-modifiers.ts
```diff
@@ -0,0 +1,193 @@
+import { ESLintUtils } from "@typescript-eslint/utils";
+import * as ts from "typescript";
+import { createRule } from "../util";
+
+const rule = createRule({
```
jakebailey, line 5: Note that this has to be added to the preset as well in the other file, now that we're doing that.

sandersn (approved):
I think it's OK to leave the TS-centric code for now. I don't believe we've had trouble with the tslint rule unlike some others that have broken with TS API changes over the years.

PR #681, opened 2023-05-21
Converted npm-naming from TSLint to ESLint

JoshuaKGoldberg:
Continues #648 by migrating `npm-naming` from TSLint to ESLint.

~~...mostly. This removes the dependency on `tslint` but still operates on the TS tree instead of TSESTree tree. I ran out of energy overhauling these more TypeScript-intertwined rules... is this ok as an intermediary step?~~ I ended up overhauling this to go all the way to using just ESLint/ESTree. ✔️ 

Also updates the `dts-critic` script's hardcoded naming rule modifiers to operate on `.eslintrc.json` instead of `tslint.json`.


packages/dtslint/src/rules/npm-naming.ts
```diff
@@ -0,0 +1,234 @@
...
+  | "NeedsExportEquals"
+  | "NoDefaultExport"
+  | "JsPropertyNotInDts"
+  | "DtsPropertyNotInJs"
+  | "JsSignatureNotInDts"
+  | "DtsSignatureNotInJs";
+
+const messageMultiLine = `{{ explanation }}
+If you won't fix this error now or you think this error is wrong,
+you can disable this check by adding the following options to your project's tslint.json file under "rules":
```
sandersn, line 38: tslint.json or .eslintrc.json?

packages/dtslint/src/rules/npm-naming.ts
```diff
@@ -0,0 +1,234 @@
...
+      try {
+        const optionsWithSuggestions = toOptionsWithSuggestions(options);
+        const diagnostics = critic(sourceFile.fileName, /* sourcePath */ undefined, optionsWithSuggestions);
+        const errors = filterErrors(diagnostics);
+        for (const error of errors) {
+          switch (error.kind) {
+            case ErrorKind.NoMatchingNpmPackage:
+            case ErrorKind.NoMatchingNpmVersion:
+            case ErrorKind.NonNpmHasMatchingPackage:
+              lookFor("// Type definitions for", error);
```
sandersn, line 137: note to self: this will have to change in the monorepo branch since DT headers will have moved to package.json (although I think this is just intended to put the error at the beginning of the file).

sandersn (changes requested):
What are the drawbacks of using the entire TS AST? Is it slower? Harder to maintain? I think of npm-naming as mostly delegating to dts-critic, but there is more code than I thought here.

packages/dtslint/src/rules/npm-naming.ts
```diff
@@ -0,0 +1,233 @@
...
+            },
+            required: ["mode"],
+          },
+        ],
+      },
+    ],
+  },
+  // TODO: This code is a modified version of the old TSLint rule,
+  // and still primarily uses TypeScript nodes. It would be better
+  // to switch it to using TSESTree nodes like other ESLint rules.
```
JoshuaKGoldberg, line 112: > What are the drawbacks of using the entire TS AST? Is it slower? Harder to maintain?

There actually aren't that many drawbacks.
* It's not slower; `@typescript-eslint/parser` always creates both the ES tree and TS tree anyway. 
* It might be harder to maintain: ESLint rule authors are typically used to working on the ES tree

packages/eslint-plugin/test/npm-naming.test.ts
JoshuaKGoldberg, line 1: I do see the new snapshot setup added... but these tests were working before and I'm much more familiar with the standard rule tester. 🤷

jakebailey: Yes, I would use the old style for this and then disable the rule within the fixtures dir, since they're all fake and wouldn't work anyway.

packages/eslint-plugin/src/util.ts
```diff
@@ -116,3 +116,24 @@ export function getImportSource(
...
+  }
+
+  if (path.basename(fileName) !== "index.d.ts") {
+    return false;
+  }
+
+  let parent = path.dirname(fileName);
+  // May be a directory for an older version, e.g. `v0`.
+  // Note a types redirect `foo/ts3.1` should not have its own header.
+  if (allowNested && /^v(0\.)?\d+$/.test(path.basename(parent))) {
```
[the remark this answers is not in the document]
JoshuaKGoldberg: I'm not quite confident enough in this repo to do that myself, but I do like the idea. 🙂

packages/dtslint/src/index.ts
```diff
@@ -21,6 +21,7 @@ async function main(): Promise<void> {
   let dirPath = process.cwd();
   let onlyTestTsNext = false;
   let expectOnly = false;
+  let npmNamingOnly = false;
```
jakebailey, line 24: Ah, I think I should have described this better. This check should still be non-optional and not have a flag; it's just that it should be enabled via the overrides passed to the ESLint class, rather than enabled via a preset. No need to make a flag or anything.

packages/dtslint/src/lint.ts
```diff
@@ -27,8 +28,10 @@ export async function lint(
   // TODO: To remove tslint, replace this with a ts.createProgram (probably)
   const lintProgram = Linter.createProgram(tsconfigPath);
 
-  // tslint no longer checks ExpectType; skip linting entirely if we're only checking ExpectType.
-  const linter = !expectOnly ? new Linter({ fix: false, formatter: "stylish" }, lintProgram) : undefined;
+  // tslint no longer checks ExpectType or npm-naming; skip linting entirely if we're only checking either.
+  // TODO: soon we shall remove tslint altogether!
+  const linter =
+    !expectOnly && !npmNamingOnly ? new Linter({ fix: false, formatter: "stylish" }, lintProgram) : undefined;
```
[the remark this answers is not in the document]
jakebailey: This is alright to disable tslint, however, I meant that below in this file there's some code that builds the options to pass to ESLint, and we should tack on the rule there. (I'd comment on that line but I can't on github)

packages/dtslint/src/updateConfig.ts
```diff
@@ -15,7 +15,6 @@ import { Configuration as Config, ILinterOptions, IRuleFailureJson, Linter, Lint
 import * as ts from "typescript";
```
jakebailey, line 15: Is this script needed anymore? Does it work without tslint?

jakebailey: I guess we aren't removing tslint in this PR? That's fine too. Cleanup can be later.

packages/eslint-plugin/src/rules/index.ts
```diff
@@ -37,5 +38,6 @@ export const rules = {
   "no-single-declare-module": noSingleDeclareModule,
   "no-old-dt-header": noOldDTHeader,
   "no-import-of-dev-dependencies": noImportOfDevDependencies,
+  "npm-naming": npmNaming,
```
jakebailey, line 41: I can't comment on the file since it wasn't modified, but the actual preset that's exported from this package should filter this entry out of the rules object so that it's not enabled by default.

packages/dtslint/src/updateConfig.ts
JoshuaKGoldberg: Yeah my strategy guess was to get this one in minimally and then do a full TSLint removal. That'd be my preference. But I'm flexible :)

packages/eslint-plugin/test/__snapshots__/plugin.test.ts.snap
```diff
@@ -0,0 +1,655 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`plugin should have the expected exports 1`] = `
+{
+  "configs": {
```
jakebailey, line 5: This is a new snapshot I added to the PR to baseline our exported plugin; you'll see in the commit that filtered out `npm-naming` that it was removed (but the rule remained as a rule).

jakebailey:
This PR should now be cleaned up and ready for review/merge (plus a change in DT itself to move the linting over).

andrewbranch (approved):

jakebailey:
Running locally, I get:

```
4> alpha-shape failing:
4> TypeError: Error while loading rule '@definitelytyped/npm-naming': versions.forEach is not a function
4> Occurred while linting /home/jabaile/work/DefinitelyTyped/types/alpha-shape/index.d.ts
4>     at Object.maxSatisfying (/home/jabaile/work/DefinitelyTyped-tools/node_modules/.pnpm/semver@7.5.4/node_modules/semver/ranges/max-satisfying.js:13:12)
4>     at getMatchingVersion (/home/jabaile/work/DefinitelyTyped-tools/packages/dts-critic/dist/index.js:273:40)
4>     at checkNpm (/home/jabaile/work/DefinitelyTyped-tools/packages/dts-critic/dist/index.js:247:24)
4>     at dtsCritic (/home/jabaile/work/DefinitelyTyped-tools/packages/dts-critic/dist/index.js:110:28)
4>     at create (/home/jabaile/work/DefinitelyTyped-tools/packages/eslint-plugin/dist/rules/npm-naming.js:120:56)
4>     at Object.create (/home/jabaile/work/DefinitelyTyped-tools/node_modules/.pnpm/@typescript-eslint+utils@6.11.0_eslint@8.53.0_typescript@5.2.2/node_modules/@typescript-eslint/utils/dist/eslint-utils/RuleCreator.js:38:20)
4>     at createRuleListeners (/home/jabaile/work/DefinitelyTyped-tools/node_modules/.pnpm/eslint@8.53.0/node_modules/eslint/lib/linter/linter.js:910:21)
4>     at /home/jabaile/work/DefinitelyTyped-tools/node_modules/.pnpm/eslint@8.53.0/node_modules/eslint/lib/linter/linter.js:1081:110
4>     at Array.forEach (<anonymous>)
4>     at runRules (/home/jabaile/work/DefinitelyTyped-tools/node_modules/.pnpm/eslint@8.53.0/node_modules/eslint/lib/linter/linter.js:1018:34)
```

jakebailey:
Oh npm

```console
$ npm info alpha-shape --json --silent versions dist-tags
{
  "versions": "1.0.0",
  "dist-tags": {
    "latest": "1.0.0"
  }
}
```

packages/dtslint/src/lint.ts
```diff
@@ -27,8 +27,8 @@ export async function lint(
   // TODO: To remove tslint, replace this with a ts.createProgram (probably)
   const lintProgram = Linter.createProgram(tsconfigPath);
```
jakebailey, line 28: We only use this program to build a list of files; not for this PR, but we're going to have to remove this when we remove tslint so I expect that we will just ask eslint to run on all non-ignored files (which is a lot better than having to find a list of all files here).

jakebailey: That being said, since we run type-using rules and also use project=true, it's likely that people will have random files that _aren't_ in the program that may break via ts-eslint complaining about not being in a program, or via new lints that were previously uncaught.

JoshuaKGoldberg: Cost reduction: `Linter.createProgram` is a standalone `public static`, so an equivalent could always be re-written.

jakebailey: Probably, yeah, though it seems a bit overkill to build a Program to only use it to list files!

PR #687, opened 2023-06-28
cmake-js added as external library without any types

MDReal32:


sandersn:
What PR on Definitely Typed is this for?

Also, cmake-js doesn't appear to have types in its package, so I don't know of a reason to allow it as a package.json dependency on Definitely Typed. (But seeing a real PR might make it clear to me.)

MDReal32:
Sorry. That was a mistake from my side. I didn't correctly read a exception message. Closing this pr.

PR #697, opened 2023-07-30
feat: Added aggregate-error to allowed dependencies

seebeen:
Aggregate error is a package uses by semantic-release, and is required for the type definitions for v21.

https://www.npmjs.com/package/aggregate-error

PR on DT: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66223


seebeen:
Thanks for the quick response1

PR #699, opened 2023-07-31
feat: remove formatting rules rules from dtslint

JoshuaKGoldberg:
Targets a bit of both of:

* https://github.com/microsoft/DefinitelyTyped-tools/issues/648: by removing some TSLint rules altogether
* https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/65993: by starting on step 2 of the plan


packages/dtslint/dtslint.json
```diff
@@ -14,33 +13,12 @@
 
         "comment-format": [true, "check-space"], // But not check-uppercase or check-lowercase
         "interface-name": [true, "never-prefix"],
-        "max-line-length": [true, 200],
```
jakebailey, line 17: Kinda funky that we have 200 here but 120 in the prettier config (and now dprint config). Of course, I personally use 120 so 🤫

packages/dtslint/src/rules/noPaddingRule.ts
```diff
@@ -1,86 +0,0 @@
-import * as Lint from "tslint";
```
sandersn, line 1: aha, so that's what the other dtslint rules were doing!

sandersn (approved):
Let's merge this shortly before the first dprint PR on DT goes in.

mastrzyz:
@JoshuaKGoldberg  if I understand correctly, this should be a follow-up PR https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66514 right?

PR #701, opened 2023-08-04
Install publisher dependencies in GHA instead of on Azure Functions host

andrewbranch:


.github/workflows/deploy.yml
```diff
@@ -25,18 +25,21 @@ jobs:
...
-        enable-oryx-build: true
+        package: ${{ runner.temp }}/publisher
     - name: upload artifacts
-      if: always()
+      if: failure()
       uses: actions/upload-artifact@v1
       with:
         name: temp
-        path: /home/runner/work/_temp
+        path: ${{ runner.temp }}
```
sandersn, line 45: I guess runner.temp is already defined?

.github/workflows/deploy.yml
```diff
@@ -25,18 +25,21 @@ jobs:
...
         app-name: types-publisher
         slot-name: production
         publish-profile: ${{ secrets.TYPESPUBLISHERFUNCTIONSPUBLISHPROFILE }}
-        package: packages/publisher
-        scm-do-build-during-deployment: true
-        enable-oryx-build: true
+        package: ${{ runner.temp }}/publisher
     - name: upload artifacts
-      if: always()
+      if: failure()
```
sandersn, line 41: does `if: failure()` mean the old thing still runs as a backup?

sandersn (approved):
Looks worth a try, although I have a couple of questions.

.github/workflows/deploy.yml
andrewbranch: We were only doing this for debugging purposes, and you typically don't need to debug unless something goes wrong. The upload takes much longer now that the temp dir has a copy of the app in it.

.github/workflows/deploy.yml
andrewbranch: Yes. I had a relevant Actions docs link open but it’s gone forever now 😅

RyanCavanaugh (approved):

PR #703, opened 2023-08-04
Support 5.3

sandersn:
To be merged after 5.2 has its own branch and `typescript@next`'s version is 5.3


sandersn:
Oops, I merged this too early. It should go in after 5.3 is actually available on npm, probably Tuesday. I'll revert it for now.

PR #706, opened 2023-08-07
Move ESLint rules into a plugin

### LINKED ISSUE #691: Don’t use ESLint rulePaths
Currently `dtslint` uses the ESLint option `rulePaths`. This registers rules that can only be found by this particular instance of ESLint. In other words: It does not integrate with how one would normally run ESLint, e.g. by using `eslint .` or in an editor.

The solution would be to make these custom rules part of a plugin.

Refs https://github.com/DefinitelyTyped/DefinitelyTyped/issues/61764

remcohaszing:
With this change, DefinitelyTyped could replace all explicit custom rules with the `@definitelytyped:all` preset. This allows users to also run ESLint via the `eslint` CLI or an editor integration.

This requires some changes to DefinitelyTyped.

1. `@definitelytypes/eslint-plugin` needs to be added as a dependency
2. The ESLint config needs to be adjusted.

Closes #691


remcohaszing:
I’m not really against that, but that would mean that if this plugin is used elsewhere, this imposes additional dependencies.

Some of these rules could be useful outside of DefinitelyTyped, e.g. `export-just-namepace` and `no-any-union`.

jakebailey:
Sure, but you don't have to extend our config to get them.

But, I do wonder if people actually do want those rules; after all, they've never been able to have them already and that's been true for a while. If they're that useful, maybe they'd be good for ts-eslint itself?

remcohaszing:
> If they're that useful, maybe they'd be good for ts-eslint itself?

My thought exactly!

packages/eslint-plugin/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "outDir": "dist",
+    "rootDir": "src"
+  },
+  "include": ["src", "src/rules"],
+  "references": [{ "path": "../header-parser" }]
```
sandersn, line 8: shouldn't packages/dtslint/tsconfig.json need to add a dependency on "../eslint-plugin"?

sandersn (changes requested):
Looks good. I have 3 questions.

- does packages/dtslint/tsconfig.json need to add a reference to ../eslint-plugin?
- do the remaining tslint rules stay in dtslint or do they also move to eslint-plugin?
- what does the associated DT PR look like? Just adding the `@definitelytyped/` prefix in .eslintrc.json?

remcohaszing:
> Looks good. I have 3 questions.
> 
> * does packages/dtslint/tsconfig.json need to add a reference to ../eslint-plugin?

No. `packages/dtslint` doesn’t import `@definitelytyped/eslint-plugin`. These packages are unrelated now.

> * do the remaining tslint rules stay in dtslint or do they also move to eslint-plugin?

TSLint rules aren’t ESLint rules, so they don’t belong in an ESLint plugin.

> * what does the associated DT PR look like? Just adding the `@definitelytyped/` prefix in .eslintrc.json?

Adding the `@definitelytyped/` prefix is one option. I also added a preset which can be used instead. So the diff could look something like this:

```diff
      plugins: ['@typescript-eslint', 'jsdoc'],
+     extends: ['@definitelytyped'],
      rules: {
-         'dt-header': 'error',
-         'export-just-namespace': 'error',
# etc
```

Note that the preset I added uses [`overrides`](https://eslint.org/docs/latest/use/configure/configuration-files#how-do-overrides-work), but the original config in DefinitelyTyped does not. This means these rules only apply to TypeScript files, not JavaScript files. It also means it’s possible to run `eslint` on a directory, i.e. `eslint .` or `eslint types/react`.

Also some files may have disabled these rules. Their `eslint-disable` comments must be patched, which can be done via a simple search and replace.

jakebailey:
We chatted about this yesterday afternoon and I think we definitely want the preset in the plugin.

The other answers make sense to me!

remcohaszing:
Do you mean the preset enabling all `@definitelytyped` rules? In that case it’s ready as-is.

Or did you mean to add a preset to enable all rules currently in [DefinitelyTyped’s `.eslintrc`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/.eslintrc.cjs)?

jakebailey:
Both, so we can just have an extends in the DT repo and that's it; only one source of truth.

packages/eslint-plugin/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "dependencies": {
+    "@definitelytyped/header-parser": "^0.0.166-next.1",
+    "@typescript-eslint/types": "^5.56.0",
+    "@typescript-eslint/utils": "^5.55.0"
+  },
+  "peerDependencies": {
+    "@typescript-eslint/eslint-plugin": "5",
+    "@typescript-eslint/parser": "5",
+    "eslint": "8",
+    "eslint-plugin-jsdoc": "44",
```
jakebailey, line 33: These qualifiers look out of place; the other packages use carat notation. Probably they should match the monorepo?

remcohaszing: Other peer dependencies in this repo don’t match either format, but I changed these to use caret ranges.

jakebailey: Dang! Mainly, I was just looking for the minimums here to look like the other packages, but I guess this monorepo isn't consistent?

remcohaszing: It’s not really inconsistent. Other packages only have one of these in their peer dependencies:

```jsonc
"typescript": ">= 3.0.0-dev || >= 3.1.0-dev || >= 3.2.0-dev || >= 3.3.0-dev || >= 3.4.0-dev || >= 3.5.0-dev || >= 3.6.0-dev || >= 3.7.0-dev || >= 3.8.0-dev || >= 3.9.0-dev || >= 4.0.0-dev || >=5.0.0-dev"
```

```jsonc
"typescript": "*"
```

So adding these new peer dependencies is basically a new concept for this monorepo.

jakebailey: To be clear, I wasn't meaning only peer dependencies, I meant specifically that we may want to copy the version strings from this file: https://github.com/microsoft/DefinitelyTyped-tools/blob/4ff7996405e7b3a61ae47bbbbaca3f86f492b012/packages/dtslint/package.json#L29

sandersn (approved):
This looks good to me; I still need to test it on DT tomorrow morning before merging.

remcohaszing:
I reverted the lockfile changes (it still matches). Now `npm test` passes. I also synced the version of `@definitelytypes/eslint-plugin` with the rest of the packages.

sandersn:
I've got it running locally and have found that all the rule exemptions need to have their names updated. I'll run some scripts to do that and send the PR when it's ready. It'll be Monday before it's done though.

sandersn:
@JoshuaKGoldberg do you have an idea why code like this would error with "Type parameter 'Variable' is **never** used"? `getSoleUse` in noUnnecessaryGenerics relies on symbol equality, so my best guess is (1) checker.getSymbolAtLocation now fails at one of these locations or (2) node.forEachChild is broken in some way that prevents the rule from seeing `Variables` inside `Notifier<Variables>`.

I'll try debugging this on Monday but I thought you might have an idea off the top of your head, since it feels like a tree mismatch problem.

```ts
/** Observes given notifier using the provided observer */
export function observe<Variables = object>(
    absintheSocket: AbsintheSocket,
    notifier: Notifier<Variables>,
    observer: Observer<Variables>,
): AbsintheSocket;
```

JoshuaKGoldberg:
Hmm, no clue. If you run out of leads let me know and I can take a look.

sandersn:
`ts.isIdentifier` never returns true, and when I dumped the syntax kinds from the tree walk, they were completely wrong. So there's some version mismatch of typescript somewhere.

remcohaszing:
Is it related to this PR though? I’m not sure how you’re testing this, but this PR doesn’t change any dependencies, whereas DefinitelyTyped doesn’t have a lockfile. Possibly an old `node_modules` folder could affect your testing.

Also there are two versions of the [various `@typescript-eslint` package versions in this package’s lockfile](https://github.com/microsoft/DefinitelyTyped-tools/blob/master/yarn.lock#L2307-L2313). Unlike `package-lock.json`, `yarn.lock` doesn’t really lock dependencies in the sense that it specifies which dependencies are installed where. It only specifies that given a version range, it should resolve to an exact version.

sandersn:
It might not be related to this PR, it's true. Definitely Typed specifies `typescript@next` as its version of typescript, in fact. I don't know how ts-eslint resolves its typescript but it could be 5.3-dev. The version that eslint-plugin resolves is 5.0, though, because I printed it to the console while no-unnecessary-generics was running. I don't know how to do the same for the typescript that ts-eslint is using.

Updating devDeps of typescript from 5.0 to 5.1 fixes the problem though. I'm going to try a run with that and push a commit if it succeeds.

sandersn:
Well, that breaks `void-return`, so I think the versions are still different.

sandersn:
I'm going to merge this and then send my DT PR and hope that the version mixup is due to my use of `npm link` for testing.

sandersn:
@remcohaszing would you mind looking at the DT fixup PR at https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66439

sandersn:
This is reverted now; it looks like the version mismatch was a problem on CI too.

The reason that's a problem is that tslint rules rely on the deprecated property originalKeywordKind which was removed in 5.0 or later (I can't find the post mentioning it.)

One way to solve this is to completely get rid of tslint; we're close to doing that with the dprint switch and the PRs that @JOshuaKGoldberg has provided. The only one missing is a switch to the community version of ExpectType. So I'll try to allocate time to making that switch happen as I have time for it.

PR #711, opened 2023-08-22
Less parallelism in retag

sandersn:
Too much: It hits the npm rate limit and can't tag anymore.
Too little: It hits the Actions time limit and gets killed.

I hope 5 is Just Right, but it might need some tweaking.

### DIFF
--- a/packages/retag/src/index.ts
+++ b/packages/retag/src/index.ts
@@ -14,6 +14,7 @@ import {
   loggerWithErrors,
   LoggerWithErrors,
   cacheDir,
+  nAtATime,
 } from "@definitelytyped/utils";
 import {
   AnyPackage,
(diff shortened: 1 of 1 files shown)


sandersn:
Weird CI failure: it says that yarn.lock points to an old version of `@definitelytyped/header-parser` and `typescript-versions` -- 0.0.166 instead of 0.0.168. It claims the culprit is individual package.jsons that point to 166 instead of 168, but I don't see any. yarn.lock has a couple of 166 although I don't know how they got there.

I'll investigate more tomorrow.

sandersn:
Missed it yesterday, but eslint-plugin's dependency on header-parser is still 166. Updating that should fix the problem.

PR #712, opened 2023-08-23
add `xlsx` package for `tabulator-tables` types

tetarchus:
Adds `xlsx` package (SheetJS) for inclusion in tabulator-tables types.

This package is used within Tabulator, and options can be passed in as part of Tabulator's own [XLSX Downloader Options](https://tabulator.info/docs/5.5/release#download).

The package is only available via CDN currently (not NPM) according to their [documentation](https://docs.sheetjs.com/docs/getting-started/installation/nodejs).

### DIFF
--- a/packages/definitions-parser/allowedPackageJsonDependencies.txt
+++ b/packages/definitions-parser/allowedPackageJsonDependencies.txt
@@ -873,6 +873,7 @@ web-tree-sitter
 winston
 winston-transport
 workbox-build
+xlsx
 xmlbuilder
 xpath
 yup


sandersn:
Is there an associated Definitely Typed PR?

sandersn:
Also, how does allowing a reference to the `xlsx` package on npm help if the current version isn't on npm? Are the tabulator-tables types going to rely on the 0.18 types that are still on npm? Are they going to point to a github repo?

tetarchus:
Apologies, I was finishing off the PR in the main DT repo: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66458

As the CDN is kept up to date, I have used that in the `package.json` of the `tabulator-tables` type package.
I'm assuming that it will still resolve fine as the name of the package still shows as `xlsx` in the `package.json`, however if there is another preferred method of using these types of packages please let me know.

tetarchus:
This is the suggested method of installing the dependency via the `package.json`. If that's the case then I think there are 3 options, and not really sure on the best one here:

- Use the npm version, which is currently 2 minor versions behind, and if they're not longer publishing there, then it will only potentially get more behind.
- Add a new `@types/xlsx` package to DT, however the package provides its own types, so this would be largely redundant, other than to provide values to `tabulator-tables`, and would need maintaining separately and cause confusion.
- Copy over the types from their repo into the `index.d.ts` of `tabulator-tables`. There are 2 types that are actually used (`WritingOptions`, which extends `CommonOptions`) found [here](https://git.sheetjs.com/sheetjs/sheetjs/src/commit/485a4f30f17ee6e2d79680133251f0dbfff74c6d/types/index.d.ts#L256), however this would then mean that the `@types/tabulator-tables` package may need updating if/when `xlsx` is updated as well.

I would lean towards the last option personally, as the chances that those types change is smaller than all of their types changing, but not sure if this is the best option.

tetarchus:
Closed as no longer required

PR #716, opened 2023-08-24
Enabled @typescript-eslint/prefer-namespace-keyword

JoshuaKGoldberg:
Corresponds to https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66292 -> https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66292#discussion_r1303614848.

### DIFF
--- a/packages/eslint-plugin/src/rules/index.ts
+++ b/packages/eslint-plugin/src/rules/index.ts
@@ -16,7 +16,6 @@ import * as redundantUndefined from "./redundant-undefined";
 import * as trimFile from "./trim-file";
 
 export const rules = {
-  "@typescript-eslint/prefer-namespace-keyword": "error",
   "dt-header": dtHeader,
   "export-just-namespace": exportJustNamespace,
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/src/rules/index.ts
```diff
@@ -16,6 +16,7 @@ import * as redundantUndefined from "./redundant-undefined";
 import * as trimFile from "./trim-file";
 
 export const rules = {
+  "@typescript-eslint/prefer-namespace-keyword": "error",
```
jakebailey, line 19: Did you mean to stick this into the preset file? This is the rules object for the plugin.

JoshuaKGoldberg: Oh, yes, thank you

sandersn:
This should only be merged and shipped once https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66292 is also ready to merge.

PR #718, opened 2023-08-30
chore: Audit unused eslint disables.

mastrzyz:
# What?

Fail the CI/Local build if we are using eslint-disable spuriously.

More information -> [here](https://eslint.org/docs/latest/use/command-line-interface#:~:text=%2D%2Dreport%2Dunused%2Ddisable%2Ddirectives,-This%20option%20causes&text=This%20can%20be%20useful%20to,which%20are%20no%20longer%20applicable.&text=When%20using%20this%20option%2C%20it,or%20custom%20rules%20are%20upgraded.)

# Why?

It's a bit confusing seeing `eslint-disable` on some of the violations like the `@ts-ignores` and some not.

This either was due to accidentally removing that rule or we no longer want to enforce, if the latter, lets clean it up.

# How?

Unfortunately, eslint **only** errors if we have this as part of the CLI Arguments `--report-unused-disable-directives`.

```
/Users/marcinstrzyz/src/DefinitelyTyped-tools/packages/dtslint/src/rules/noRedundantJsdoc2Rule.ts
   81:7  error  Unused eslint-disable directive (no problems were reported from '@typescript-eslint/ban-ts-comment')
  164:5  error  Unused eslint-disable directive (no problems were reported from '@typescript-eslint/ban-ts-comment')
```

adding `reportUnusedDisableDirectives` to eslintrc only provides a warning.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "scripts": {
     "prepare": "lerna bootstrap --no-ci",
     "pretest": "yarn build",
-    "lint": "eslint . --ext .ts",
+    "lint": "eslint . --report-unused-disable-directives --ext .ts",
     "format": "prettier --write 'packages/**/*.ts'",
     "test": "jest",
     "build": "tsc -b .",
--- a/packages/dtslint/src/lint.ts
+++ b/packages/dtslint/src/lint.ts
@@ -156,7 +156,6 @@ interface Err {
 }
 function testNoLintDisables(disabler: "tslint:disable" | "eslint-disable", text: string): Err | undefined {
   let lastIndex = 0;
-  // eslint-disable-next-line no-constant-condition
   while (true) {
     const pos = text.indexOf(disabler, lastIndex);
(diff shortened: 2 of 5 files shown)


jakebailey:
This seems good though I think even better would be for you to set this in dtslint itself so it applies to DefinitelyTyped!

mastrzyz:
> This seems good though I think even better would be for you to set this in dtslint itself so it applies to DefinitelyTyped!

Hmm , wouldn’t that cause a large influx of potential new unused Eslint disables if we upgrade the package there ? Happy to do that but think we would need a separate PR since it will be large

jakebailey:
It'd cause failures on DT, yes, but they'd only be good failures we'd want to fix.

sandersn (approved):

sandersn:
Let's take this separately, but I agree that having this on DT is even more useful than here.

jakebailey (approved):

PR #731, opened 2023-09-13
Migrate types-publisher to GitHub Actions

andrewbranch:


.github/workflows/publish-packages.yml
```diff
@@ -0,0 +1,27 @@
+name: Publish Packages
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '*/30 * * * *' # every 30 minutes
+
+# publishing can take longer than 30 minutes but we never want to
+# run more than one publish job at a time
+concurrency: 1
```
jakebailey, line 10: FWIW this might be misleading; this option is effectively the name of the lock, not a number.

andrewbranch: I assumed it was one or the other and the result would be the same 😆

.github/workflows/publish-packages.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+      - run: yarn install --frozen-lockfile
+      - run: yarn build
+      - uses: actions/cache@v3
+        with:
+          path: ./cache
+          key: always
```
jakebailey, line 23: This unfortunately won't do a good thing; it'll save the cache once, but since caches can never be overwritten, it'll keep downloading the first cached item and never update it, and it'll never die either because it's still being referenced.

The only way to have a cache which gets periodically deleted is to give it a new name.

Depending on how long you want things to be cached, you can do something like I'm planning to do on DT for pnpm so we get _some_ caching while still getting latest: https://github.com/jakebailey/DefinitelyTyped/blob/1666921ac1474b381720a3535a642c64d91d28ae/.github/workflows/CI.yml#L47

TL;DR is to basically construct a cache key that encodes the current time such that you'll get the most recent one, and just rely on GitHub deleting the old ones once they time out or the max cache size is hit (which deletes the oldest entries).

andrewbranch:
I think we just have to merge this to try it out

PR #733, opened 2023-09-14
Migrate from yarn and lerna to pnpm and changesets

andrewbranch:
I assume I’m forgetting something


package.json
```diff
@@ -23,11 +22,17 @@
     "eslint-plugin-import": "^2.22.1",
     "eslint-plugin-jsdoc": "^30.7.8",
     "jest": "^29.5.0",
-    "lerna": "^4.0.0",
     "prettier": "^2.6.2",
     "ts-jest": "^29.0.5",
     "tslint": "^6.1.2",
     "tslint-microsoft-contrib": "^6.2.0",
     "typescript": "^5.0.2"
+  },
```
jakebailey, line 30: Can you set something like `"packageManager": "pnpm@8.7.4"` here?

.github/workflows/update-ts-version-tags.yml
```diff
@@ -15,19 +15,24 @@ on:
...
   publish:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
     - uses: actions/setup-node@v2
       with:
-        node-version: '14'
+        node-version: 'lts/*'
+    - run: npm install -g pnpm
+    - run: pnpm config set store-dir $PNPM_CACHE_FOLD
```
jakebailey, line 31: FWIW there's also an action-y way to do this which may be faster: https://github.com/jakebailey/esbuild-playground/blob/2149487e8726b41ecac8ae95e2c50706b9047dca/.github/workflows/gh-pages.yml#L37

Or just setting `cache: pnpm` on the setup-node action, though I'm pretty sure that doesn't work because pnpm's not installed in time.

.github/workflows/update-ts-version-tags.yml
```diff
@@ -15,19 +15,24 @@ on:
     # https://crontab.guru/#5_8_*_*_1
     - cron: "5 8 * * 1"
 
+env:
+  CI: true
```
jakebailey, line 19: Hm, why was this needed? GHA already should set this to `CI=1`.

andrewbranch:
I think I’m planning not to install changesets-bot since it will prompt people updating allowedPackageJsonDependencies to add a changeset, which we don’t need. It didn’t look like there’s a way to filter paths for the bot.

jakebailey:
This fails to install on my machine:

```
 WARN  Failed to create bin at /home/jabaile/work/DefinitelyTyped-tools/packages/dtslint-runner/node_modules/.bin/dtslint. ENOENT: no such file or directory, open '/home/jabaile/work/DefinitelyTyped-tools/packages/dtslint-runner/node_modules/@definitelytyped/dtslint/dist/index.js'
 WARN  Failed to create bin at /home/jabaile/work/DefinitelyTyped-tools/packages/publisher/node_modules/.bin/retag. ENOENT: no such file or directory, open '/home/jabaile/work/DefinitelyTyped-tools/packages/publisher/node_modules/@definitelytyped/retag/dist/retag.js'
Done in 2.1s
```

We probably need to be indirecting our `bins` through some script, as pnpm is going to try and create syminks.

andrewbranch:
Well it looks like packages didn't link to each other at all in CI? Everything is working on my machine so I'm confused

jakebailey:
Also, you probably want to set all of the inter-monorepo workspace deps to be `workspace:^` so they're reoslved locally.

andrewbranch:
> Failed to create bin

It’s trying to do this on install? The file doesn’t exist until `pnpm build`

jakebailey:
Right, but pnpm doesn't know that, it pulls packages from npm normally which are fully built and so are linkable instantly. `build` is just a random script name to it and it's not going to run preinstall or anything. It's "a" reason to have a `bin` dir which just require/import's the actual file.

.github/workflows/update-ts-version-tags.yml
andrewbranch: That's what I figured, but pnpm's guide had it and I copied it

.github/workflows/update-ts-version-tags.yml
andrewbranch: this also came from pnpm's guide 🤷

jakebailey: weird, must be a different guide from https://github.com/pnpm/action-setup

andrewbranch: https://pnpm.io/using-changesets

azure-pipelines.yml
```diff
@@ -22,7 +22,8 @@ steps:
   displayName: 'guardian export'
```
jakebailey, line 22: What is this pipeline doing anymore?

packages/eslint-plugin/tsconfig.json
```diff
@@ -1,6 +1,8 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
+    "module": "nodenext",
+    "moduleResolution": "nodenext",
```
jakebailey, line 5: Not required for this PR I don't think, but should this be getting set for the entire repo via the base?

pnpm-workspace.yaml
```diff
@@ -0,0 +1,3 @@
+prefer-workspace-packages: true
+packages:
+  - 'packages/**'
```
jakebailey, line 3: You may want to set `packages/*` instead to make pnpm faster; otherwise it's going to attempt to recurse.

pnpm-workspace.yaml
```diff
@@ -0,0 +1,3 @@
+prefer-workspace-packages: true
```
jakebailey, line 1: This setting has to go into `.npmrc`; this file can only contain the package list.

packages/dtslint-runner/package.json
```diff
@@ -36,6 +36,7 @@
   },
   "devDependencies": {
     "@types/fs-extra": "^8.1.0",
+    "@types/glob": "^8.1.0",
```
jakebailey, line 39: Wrong types version here; should be ^7 though it does compile.

jakebailey (approved):
LGTM though I still get the bin warnings locally (plus see my other comments).

azure-pipelines.yml
andrewbranch: It was a thing that @sandersn added for some compliance thing

packages/eslint-plugin/tsconfig.json
andrewbranch: Yeah, it's a big to-do; I thought this might help with the declaration emit errors, but it didn’t. This could be reverted but I'll leave it and follow up

andrewbranch:
Same, I don't think it matter locally though

jakebailey (approved):

azure-pipelines.yml
sandersn: My memory is that it's a component governance used for checking for supply-chain attacks. I can't remember if it ends up also checking DT, but my memory is that it does because of the way we run our tests. Useful to have running.

PR #736, opened 2023-09-20
Retry npm install on failure in dtslint-runner

jakebailey:
I am getting so massively tired of DT runs failing on PRs due to npm install.

This PR adds a silly retry loop as I'm pretty sure this is a transient npm problem made worse by the fact that we invoke npm a billion times.

### DIFF
--- a/.changeset/early-yaks-explode.md
+++ b/.changeset/early-yaks-explode.md
@@ -1,0 +1,5 @@
+---
+"@definitelytyped/dtslint-runner": patch
+---
+
+Retry npm installs
--- a/packages/dtslint-runner/src/prepareAffectedPackages.ts
+++ b/packages/dtslint-runner/src/prepareAffectedPackages.ts
@@ -6,7 +6,7 @@ import {
   allDependencies,
   TypingsData,
 } from "@definitelytyped/definitions-parser";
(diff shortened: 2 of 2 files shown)


.changeset/early-yaks-explode.md
```diff
@@ -0,0 +1,5 @@
+---
+"@definitelytyped/dtslint-runner": patch
```
jakebailey, line 2: I'm not sure if I should actually do this; don't we bump everything at once? So maybe I shouldn't do this at all and we're just using changesets to bump/publish?

sandersn: I have no idea what this file is or what it does. Can you explain?

(First thing I want to know is: when would early yaks explode? why? what makes them early?)

jakebailey: Maybe you missed #733 which migrated things; this is a versioning mechanism and it's auto-generated to inform monorepo version bumps. https://github.com/changesets/changesets

jakebailey: Confirmed with @andrewbranch that we're going to try changeset version bumps (which will cause the monorepo to not be versioned all at once). Easy to undo later if we want. Of course, the DT PR checks use `master` so I don't think I actually need to release this ASAP but it might be good to test.

PR #738, opened 2023-09-21
Tag during publish in changeset action

jakebailey:
This has to be done manually when not using `changeset publish`.

While here, drop the env var from the format step; it's preserved from checkout and isn't used there.

### DIFF
--- a/.github/workflows/version-or-publish.yml
+++ b/.github/workflows/version-or-publish.yml
@@ -44,11 +44,9 @@ jobs:
             git commit -am "Format"
             git push
           fi
-        env:
-          GITHUB_TOKEN: ${{ secrets.TYPESCRIPT_BOT_TOKEN }}
       - uses: changesets/action@v1
         with:
-          publish: pnpm ci:publish
+          publish: pnpm changeset tag && pnpm publish -r
         env:
           GITHUB_TOKEN: ${{ secrets.TYPESCRIPT_BOT_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/package.json
+++ b/package.json
@@ -7,8 +7,7 @@
     "format": "prettier --write 'packages/**/*.ts'",
     "test": "jest",
     "build": "tsc -b .",
-    "retag": "node packages/retag/dist/retag.js",
-    "ci:publish": "pnpm publish -r"
+    "retag": "node packages/retag/dist/retag.js"
   },
   "devDependencies": {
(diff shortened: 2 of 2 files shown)


andrewbranch (approved):
Works for me, but I’m curious what `pnpm publish -r` is doing over just `changeset publish`

jakebailey:
Yeah, I honestly don't know the answer there. I'm sure @Andarist knows 😄

.github/workflows/version-or-publish.yml
```diff
@@ -44,11 +44,9 @@ jobs:
             git commit -am "Format"
             git push
           fi
-        env:
-          GITHUB_TOKEN: ${{ secrets.TYPESCRIPT_BOT_TOKEN }}
       - uses: changesets/action@v1
         with:
-          publish: pnpm ci:publish
```
andrewbranch, line 51: We should delete this package.json script if we're not going to use it

jakebailey: Oops, I thought I committed that

.github/workflows/version-or-publish.yml
```diff
@@ -44,11 +44,9 @@ jobs:
...
+          publish: pnpm changeset tag && pnpm publish -r
```
Andarist, line 49: IIRC the introduction of `changeset tag` was primarily driven by the will to use `pnpm publish -r`. So I think this should work fine. That said - I don't think there should be a need to actually use `pnpm publish -r` over `changeset publish` in most cases - maybe there was a need for it in the past though? `pnpm publish -r` might be slightly faster though since it's a single process whereas with `changeset publish` me need to spawn `pnpm publish` for each published package separately. 

`pnpm publish -r` might do its work in a topological order (I'm not 100% sure but I think that's the case) so from that angle it's better/safer than `changeset publish` (the topological order was always something I've wanted to incorporate but never got to it).

`changeset tag` doesn't do a lot - but its stdout is important for `changesets/action`. The latter parses that output with regex (I'm sorry) to know what GitHub releases should be published.

I rechecked things just now and - if I'm reading the source code correctly - `changeset tag` needs to be able to read all the current tags. I think that they will be available locally already since you are using `actions/checkout` with `fetch-depth: 0` but perhaps it's worth rechecking this.

jakebailey: Yeah, this has `fetch-depth: 0` so this should work.

Probably I'll just leave this PR as it is, as it seems fine...

PR #745, opened 2023-09-28
Fix retag bin name

sandersn:
It's index.js now, not retag.js.

### DIFF
--- a/packages/retag/package.json
+++ b/packages/retag/package.json
@@ -33,7 +33,7 @@
   },
   "main": "dist/index.js",
   "bin": {
-    "retag": "./dist/retag.js"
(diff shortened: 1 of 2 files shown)


packages/retag/package.json
```diff
@@ -33,7 +33,7 @@
   },
   "main": "dist/index.js",
   "bin": {
-    "retag": "./dist/retag.js"
+    "retag": "./dist/index.js"
```
jakebailey, line 36: Wow, I was going to leave a comment about how weird it was that we had both main and bin for the same thing, but it actually checks if it's being invoked as main

andrewbranch (approved):

PR #748, opened 2023-10-03
Fix scripts to not execute when imported by ES module, calculate versions 10 at a time

andrewbranch:
These are prerequisites for some manual testing I'm doing on pnpm workspaces, but are beneficial separately from that effort, so I'm PRing to master.

I tested `calculate-versions` at this level of parallelism and it completed in less than 5 minutes and I wasn’t rate limited. I think it takes somewhere between 30 and 60 minutes normally.


packages/utils/src/collections.ts
```diff
@@ -198,3 +198,7 @@ export function max<T>(array: readonly T[], compare?: (a: T, b: T) => number) {
           : previousValue
       );
 }
+
+export function compact<T>(array: readonly (T | undefined)[]): T[] {
+  return array.filter((x): x is T => !!x);
```
jakebailey, line 203: Can `T` ever be falsy, or should this be `!== undefined`?

jakebailey (approved):
Does this need a changeset to get releases for each of these?

andrewbranch:
I don’t need an npm release for these, but I guess I should still add a changeset.

jakebailey:
> I don’t need an npm release for these, but I guess I should still add a changeset.

Probably not required; I honestly expect that we'll do a mass bump for pnpm anyway.

andrewbranch:
> Oh, also, do you need to path.resolve the filenames in the checks

Tested with Node 20 and they were absolute. Not sure about older versions.

sandersn (approved):

PR #752, opened 2023-10-04
Rename name-related fields

andrewbranch:
A bunch of renaming, and preserving some more info starting at the header parser. New nomenclature:

- name: `@types/ember__object`
- typesDirectoryName: `ember__object`
- libraryName: `@ember/object` (or `nonNpmDescription` for non-npm packages)
- subDirectoryPath: `ember__object/v2`

### DIFF
--- a/packages/definitions-parser/src/get-affected-packages.ts
+++ b/packages/definitions-parser/src/get-affected-packages.ts
@@ -4,11 +4,11 @@ import {
   AllPackages,
   PackageId,
   PackageBase,
-  getMangledNameForScopedPackage,
   formatDependencyVersion,
   removeTypesScope,
 } from "./packages";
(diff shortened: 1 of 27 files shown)


sandersn:
It would be nice to know what the old nomenclature was, since I have a big change in flight that will benefit from this, but also need to update all its code.

andrewbranch:
Old nomenclature:

- name: `ember__object`
- fullNpmName: `@types/ember__object`
- libraryName: Inconsistent, but was intended to be either the implementation library name or a human-readable representation of it. It wasn’t enforced to be the npm library name.
- subDirectoryPath: `ember__object/v2`

andrewbranch:
I'm regenerating publishable output to ensure nothing broke here that wasn’t covered by the tests.

PR #753, opened 2023-10-06
Pnpm workspaces: fix dtslint-runner incremental

sandersn:


packages/definitions-parser/src/get-affected-packages.ts
```diff
@@ -1,104 +1,86 @@
...
+  deletions: PackageId[],
+  definitelyTypedPath: string
+): Promise<PreparePackagesResult> {
+  const allDependents = [];
+  console.log(deletions.map((d) => d.typesDirectoryName + "@" + (d.version === "*" ? "*" : formatTypingVersion(d.version))));
+  const filters = [`--filter '...[jakebailey/pnpm-workspaces-working]'`];
+  for (const d of deletions) {
+    for (const dep of allPackages.allTypings()) {
+      for (const [name, version] of dep.allPackageJsonDependencies()) {
+        if ("@types/" + d.typesDirectoryName === name && (d.version === "*" || formatTypingVersion(d.version) === version)) {
```
andrewbranch, line 21: `version` here comes directly from the potential dependent’s package.json, right? We commonly have stuff like `"react": "^17"`, but `formatTypingVersion(d.version)` would be `17`. I think we need to use something like `SemVer.satisfies` here.

packages/definitions-parser/src/get-affected-packages.ts
```diff
@@ -1,104 +1,86 @@
...
-      all.add(item);
-      workList.push(item);
+export async function getAffectedPackages(
+  allPackages: AllPackages,
+  deletions: PackageId[],
+  definitelyTypedPath: string
+): Promise<PreparePackagesResult> {
+  const allDependents = [];
+  console.log(deletions.map((d) => d.typesDirectoryName + "@" + (d.version === "*" ? "*" : formatTypingVersion(d.version))));
+  const filters = [`--filter '...[jakebailey/pnpm-workspaces-working]'`];
```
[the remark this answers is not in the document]
sandersn: Uhhh, I'm going to change it now, actually. We already have a constant in settings.ts for the branch, so I'll introduce one for the remote as well.

packages/definitions-parser/src/get-affected-packages.ts
sandersn: Good catch. Just comparing strings to strings to string around here.

PR #757, opened 2023-10-17
Always convert contributor githubUsername to url

sandersn:


packages/definitions-parser/src/packages.ts
```diff
@@ -426,8 +426,10 @@ export class TypingsData extends PackageBase {
       this.data.header.nonNpmDescription ?? unmangleScopedPackage(this.typesDirectoryName) ?? this.typesDirectoryName
     );
   }
-  get contributors(): readonly Owner[] {
-    return this.data.header.owners;
+  get contributors(): readonly Contributor[] {
+    return this.data.header.owners.map((o) =>
+      "githubUsername" in o ? { name: o.name, url: `https://github.com/${o.githubUsername}` } : o
```
jakebailey, line 431: Maybe something like:

```suggestion
      { ...(o.githubUsername ? { url: `https://github.com/${o.githubUsername}` } : undefined), ...o }
```

Old packages had `githubUsername` in them so I'm betting we should keep that?

sandersn: wellll, I guess so

jakebailey: This is an awful one liner, a function with a body would probably be better

sandersn: Implemented in a more type-system-friendly way though

jakebailey (approved):

PR #759, opened 2023-10-18
Clone and install DT in publisher

jakebailey:
The other things we run don't seem to actually need the installed state so the memory FS is probably fine. So this is a targeted fix specifically for publishing.


andrewbranch (approved):
I guess I should make getting referenced files throw on the in-memory FS in my in-progress refactor. This is a scary footgun.

jakebailey:
Verified locally: 
image

image

image

PR #760, opened 2023-10-18
Cache pnpm store in publisher

jakebailey:
This follows the same pattern as the pacote cache, but for the pnpm store. The pair of the two should not be too large to hit the 10G limit so this should work.

### DIFF
--- a/.github/workflows/publish-packages.yml
+++ b/.github/workflows/publish-packages.yml
@@ -23,12 +23,19 @@ jobs:
           node-version: 'lts/*'
       - run: npm install -g pnpm
       - run: pnpm config set store-dir $PNPM_CACHE_FOLDER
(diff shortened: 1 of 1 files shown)


andrewbranch (approved):

.github/workflows/publish-packages.yml
```diff
@@ -46,4 +53,10 @@ jobs:
         if: always()
         with:
           path: ./cache
-          key: pacote-cache-${{ github.run_id }}-${{ github.run_attempt}}
+          key: pacote-cache-${{ github.run_id }}-${{ github.run_attempt }}
+
+      - run: pnpm store prune
+      - uses: actions/cache/save@v3
+        with:
```
jakebailey, line 60: Notably, not keeping the cache here if it failed.

PR #761, opened 2023-10-18
Check newly added packages

sandersn:
even if they're missing package.json

Moves gitDeletions from git.ts to get-affected-packages, sorry for the churn.
Edit: less churn, still churny.

### DIFF
--- a/packages/definitions-parser/test/get-affected-packages.test.ts
+++ b/packages/definitions-parser/test/get-affected-packages.test.ts
@@ -28,6 +28,7 @@ testo({
     const { packageNames, dependents } = getAffectedPackagesWorker(
       allPackages,
       packageOutput,
+      [],
       [dependentOutput],
       "/dt"
     );
@@ -40,6 +41,7 @@ testo({
     const { packageNames, dependents } = getAffectedPackagesWorker(
       allPackages,
       packageOutput,
+      [],
       [dependentOutput],
       "/dt"
     );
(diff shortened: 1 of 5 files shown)


packages/definitions-parser/test/git.test.ts
```diff
@@ -55,15 +55,20 @@ When removing packages, you should only delete files that are a part of removed
         { status: "D", file: "types/jest/index.d.ts" },
         { status: "D", file: "types/jest/jest-tests.d.ts" },
       ])
-    ).toEqual({ ok: jestNotNeeded });
+    ).toEqual({
+      errors: [
+        `Unexpected file added: oooooooooooops.txt
```
sandersn, line 60: this test had an error before and I incorrectly updated it when the error disappeared, also incorrectly.

jakebailey:
I have a hard time seeing it from the code, but will this complain when people work on things like `types/node/scripts`?

sandersn:
I think not (it's checking at the types/ directory level) but I'll check.

sandersn:
No, it just adds node to the list of packages for testing.

packages/definitions-parser/test/get-affected-packages.test.ts
```diff
@@ -55,9 +57,22 @@ testo({
...
     ];
-    const { packageNames } = getAffectedPackagesWorker(allPackages, packageOutput, dependentOutput, "/dt");
+    const { packageNames } = getAffectedPackagesWorker(allPackages, packageOutput, [], dependentOutput, "/dt");
     expect(packageNames).toEqual(new Set(["jquery"]));
   },
+  newPackage() {
+    const packageOutput = ``;
+    const dependentOutput = ``;
+    const { packageNames, dependents } = getAffectedPackagesWorker(
+      AllPackages.from({ ...typesData, mistake: createTypingsVersionRaw("mistake", {}, {}) }, notNeeded),
```
jakebailey, line 67: Should there be a test that only adds a file? This seems like it's creating a typings which I would think wouldn't exist? Maybe I'm not understanding.

jakebailey (approved):

packages/definitions-parser/test/get-affected-packages.test.ts
sandersn: oops, that's left over from my misunderstanding of how package parsing works. In fact new, malformed packages do *not* show up in AllPackages, so the test is wrong--it should use the same mock data that all the other tests use.

jakebailey (approved):

PR #763, opened 2023-10-19
Add semver types to dependencies in definitions-parser

jakebailey:
Otherwise you get:

```
node_modules/.pnpm/@definitelytyped+definitions-parser@0.0.179_typescript@5.3.0-dev.20231018/node_modules/@definitelytyped/definitions-parser/dist/packages.d.ts:4:25 - error TS7016: Could not find a declaration file for module 'semver'. '/home/jabaile/work/DefinitelyTyped/node_modules/.pnpm/semver@7.5.4/node_modules/semver/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/semver` if it exists or add a new declaration (.d.ts) file containing `declare module 'semver';`

4 import * as semver from "semver";
```

### DIFF
--- a/.changeset/dirty-walls-drum.md
+++ b/.changeset/dirty-walls-drum.md
@@ -1,0 +1,5 @@
+---
+"@definitelytyped/definitions-parser": patch
+---
+
+Add semver types to dependencies in definitions-parser
--- a/packages/definitions-parser/package.json
+++ b/packages/definitions-parser/package.json
@@ -23,14 +23,14 @@
     "@definitelytyped/typescript-versions": "workspace:*",
     "@definitelytyped/utils": "workspace:*",
     "@types/node": "^14.14.35",
+    "@types/semver": "^7.5.2",
     "fs-extra": "^9.1.0",
     "pacote": "^13.6.1",
     "semver": "^7.3.7"
   },
   "devDependencies": {
     "@types/fs-extra": "^9.0.8",
-    "@types/pacote": "^11.1.5",
-    "@types/semver": "^7.5.2"
+    "@types/pacote": "^11.1.5"
   },
   "peerDependencies": {
     "typescript": "*"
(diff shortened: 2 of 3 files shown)


jakebailey:
Things are also broken a different way:

```
node_modules/.pnpm/@definitelytyped+definitions-parser@0.0.180_typescript@5.3.0-dev.20231018/node_modules/@definitelytyped/definitions-parser/dist/packages.d.ts:1:10 - error TS2305: Module '"@definitelytyped/header-parser"' has no exported member 'Contributor'.

1 import { Contributor, Header, License } from "@definitelytyped/header-parser";
           ~~~~~~~~~~~
```

Some sort of missing package bump?

andrewbranch (approved):
Yeah, header-parser hasn't published this change, but definitions-parser has been published for something else since then https://github.com/microsoft/DefinitelyTyped-tools/pull/757

jakebailey:
Oh, #757 didn't include a changeset so things shipped broken. That's probably an error; should I send a PR with that or tack it on here?

andrewbranch:
Either way, or just push one to master.

We can add the changeset bot to remind PRs to include them, or their GitHub Action to enforce that they be included. The problem with both is that they can’t be filtered to a set of files, so it would pester people who are updating allowedPackageJsonDependencies.txt. I guess we could easily write our own shell script and include it in CI checks, though.

jakebailey:
To use the changeset bot, we'd have to add the App to our repo; not sure that we can do that since it's actually installed at the org level? Maybe someone else was nice enough to have done it. Would be nice if they had a version of their App that instead could run as a GHA `pull_request_target` workflow. That actually sounds pretty straightforward to implement...

jakebailey:
Ah, that's https://github.com/changesets/bot/issues/54. I guess it was an action before but was changed to an app before `pull_request_target` became a thing.

PR #764, opened 2023-10-19
Lazy `AllPackages`

andrewbranch:
Defers reading a `types/*` directory until it’s asked for, and further defers analyzing the source files of a package (to get used/unused files, check that local references are of an allowed form, and get the full file list for content hash) until it’s asked for.


andrewbranch:
I left in some types/options for parallel parsing that are currently unused, thinking I might bring back sharded processing, but with other changes we’ve talked about, I don’t think it will be necessary. I’ll remove them now.

packages/dtslint-runner/src/main.ts
```diff
@@ -48,8 +48,8 @@ export async function runDTSLint({
   const typesPath = joinPaths(definitelyTypedPath, "types");
 
   const { packageNames, dependents } = onlyRunAffectedPackages
-    ? await prepareAffectedPackages(definitelyTypedPath, nProcesses)
-    : await prepareAllPackages(definitelyTypedPath, definitelyTypedAcquisition.kind === "clone", nProcesses);
+    ? await prepareAffectedPackages(definitelyTypedPath)
```
jakebailey, line 51: The reference to `nProcesses` went away here but the variable didn't actually go away; is there more cleanup missing?

packages/publisher/package.json
```diff
@@ -34,7 +34,6 @@
     "test": "../../node_modules/.bin/jest --config ../../jest.config.js packages/publisher",
     "clean": "node -r source-map-support/register dist/clean.js",
     "get-definitely-typed": "node -r source-map-support/register dist/get-definitely-typed.js",
-    "parse": "node -r source-map-support/register dist/parse-definitions.js",
```
jakebailey, line 37: Unrelated and not for this PR, but we should really drop `source-map-support` everywhere; `node --enable-source-maps`  has been around since Node 12, stable since node 14.18.

packages/retag/src/index.ts
```diff
@@ -47,25 +41,21 @@ async function main() {
  * This shouldn't normally need to run, since we run `tagSingle` whenever we publish a package.
  * But this should be run if the way we calculate tags changes (e.g. when a new release is allowed to be tagged "latest").
  */
-async function tag(dry: boolean, nProcesses: number, name?: string) {
+async function tag(dry: boolean, _nProcesses: number, name?: string) {
```
jakebailey, line 44: Should this process stuff also go away or are we trying to keep this API?

jakebailey (commented):
I'm somewhat noticing a lot of parallel `Promise.all`s; do we need limiters to make sure we're not hitting stuff like the GitHub / npm APIs too hard now that we aren't only doing these in plain linear loops? e.g. before we'd have maybe one or two processes in CI, all going one at a time, but now we request all packages in parallel, so I'm not sure what all happens there.

packages/retag/src/index.ts
andrewbranch: I missed this one, removed most of it.

packages/dtslint-runner/src/main.ts
jakebailey: Oh, so it still runs things in parallel? I guess I assumed it all was going away but I guess it's just parallel parse.

andrewbranch:
> but now we request all packages in parallel

I don’t think we’re making any network requests in parallel; did you see a place where that’s happening? I tried this locally and noticed that it fired off ~9000 requests to download `allowedPackageJsonDependencies.txt` so I fixed that one with a quick hack, knowing we want to move it to a CI-only check later.

packages/definitions-parser/src/lib/definition-parser.ts
```diff
@@ -53,17 +41,25 @@ function formattedLibraryVersion(typingsDataRaw: TypingsDataRaw): `${number}.${n
   return `${typingsDataRaw.header.libraryMajorVersion}.${typingsDataRaw.header.libraryMinorVersion}`;
 }
 
-export async function getTypingInfo(packageName: string, dt: FS): Promise<TypingsVersionsRaw | { errors: string[] }> {
+export async function getTypingInfo(
+  packageNameOrTypesDirectoryName: string,
```
sandersn, line 45: This name is a little unwieldy now and it doesn't look to me that the semantics have changed. If some callers always pass a directory name and some callers always pass a package name, maybe it would be worthwhile to have two entry points?

jakebailey:
I didn't see anything specific; a lot of the `Promise.all`s are at a highlevel so I wasn't super sure one way or the other.

andrewbranch:
Yeah, they’re just doing file I/O

packages/definitions-parser/src/lib/definition-parser.ts
```diff
@@ -269,33 +235,71 @@ async function combineDataForAllTypesVersions(
...
+    }
+    return data;
+  });
+
+  if (errors.length) {
+    return { errors };
+  }
+
+  return [dataForRoot, ...dataForOtherTypesVersions] as FilesForSingleTypeScriptVersion[];
+  // return Array.from(
```
sandersn, line 295: this should be deleted

packages/definitions-parser/src/lib/definition-parser.ts
```diff
@@ -304,17 +308,17 @@ interface TypingDataFromIndividualTypeScriptVersion {
...
- * @param packageName Name of the outermost directory; e.g. for "node/v4" this is just "node".
+ * @param typesDirectoryName Name of the outermost directory; e.g. for "node/v4" this is just "node".
  * @param ls All file/directory names in `directory`.
  * @param fs FS rooted at the directory for this particular TS version, e.g. `types/abs/ts3.1` or `types/abs` when typescriptVersion is undefined.
  */
-function getTypingDataForSingleTypesVersion(
+function getFilesForSingleTypeScriptVersion(
   typescriptVersion: TypeScriptVersion | undefined,
-  packageName: string,
+  typesDirectoryName: string,
```
sandersn, line 317: I wonder if we should have two different string types for packages and typeDirectories...

packages/definitions-parser/src/lib/definition-parser.ts
andrewbranch: The semantics didn’t need to change, because the packageName→directoryName transform is a noop on directory names, so we can just perform it unconditionally. I changed it because I noticed that I was passing a directory name in and wondered if that was going to break something.

packages/definitions-parser/src/lib/definition-parser.ts
```diff
@@ -328,17 +332,25 @@ function getTypingDataForSingleTypesVersion(
...
+      tsconfig.files ?? [],
+      fs,
+      typesDirectoryName,
+      moduleResolutionHost,
+      compilerOptions
+    ));
+  } catch (err) {
+    if (err instanceof Error) {
+      errors.push(err.message);
+    }
```
sandersn, line 348: maybe should throw here in the else case

packages/definitions-parser/src/packages.ts
```diff
@@ -448,12 +508,30 @@ export class TypingsData extends PackageBase {
...
+    }
+    return this.typesVersionsFiles.flatMap((v) => v.declFiles);
   }
-  get dtsFiles(): readonly string[] {
-    return this.data.files.filter((f) => f.endsWith(".d.ts") || f.endsWith(".d.mts") || f.endsWith(".d.cts"));
+
+  getFiles(): readonly string[] {
+    const files = this.tryGetFiles();
+    if ("errors" in files) {
+      throw new Error(`Errors while reading package ${this.name}: ${files.errors.join("\n")}`);
```
sandersn, line 526: did this throw before? I think that it read from the (virtual) filesystem -- so what kind of errors actually get thrown here?

Mainly asking because I thought tryGetFiles would be the more-used method, but getFiles turns out to be.

packages/definitions-parser/src/packages.ts
```diff
@@ -83,71 +107,109 @@ export class AllPackages {
...
+    }
+    if (this.errors.has(typesDirectoryName)) {
+      return undefined;
+    }
+    const raw = await getTypingInfo(typesDirectoryName, this.dt);
+    if (!raw) {
+      return undefined;
+    }
+    if ("errors" in raw) {
+      this.errors.set(typesDirectoryName, raw.errors);
```
sandersn, line 156: this error handling pattern makes me want `tryGetTypingsVersion` and `errors` both to be private. Just in case.

packages/definitions-parser/src/packages.ts
```diff
@@ -83,71 +107,109 @@ export class AllPackages {
...
       }
     }
   }
+
+  private async readAllTypings() {
+    if (this.isComplete) {
+      return;
+    }
+    // populate cache so every directory doesn't try to request this from GH
+    await getAllowedPackageJsonDependencies();
```
sandersn, line 201: how does this populate a cache? I'm not seeing it in getAllowedPackageJsonDependencies

sandersn: oh I missed the `withCache` wrapping the whole thing

packages/definitions-parser/src/packages.ts
andrewbranch: These errors would have been thrown way earlier, during the parsing of all definitions. In the end, I think changing the error handling here wasn’t significant. I did it this way because I initially kept this as a `get files()` accessor, which I didn’t want to throw on access. But I changed it to a method call to ensure you couldn’t generate errors without noticing. So I think I can simplify this a bit.

packages/definitions-parser/src/packages.ts
```diff
@@ -1,80 +1,104 @@
...
-  }
-
   /** Use for `--single` tasks only. Do *not* call this in a loop! */
-  static async readSingle(name: string): Promise<TypingsData> {
-    const data = await readTypesDataFile();
-    const raw = data[name];
-    if (!raw) {
+  static async readSingle(dt: FS, name: string): Promise<TypingsData> {
+    const data = await getTypingInfo(name, dt);
+    if (!data) {
```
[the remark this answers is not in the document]
andrewbranch: This was motivated by the `allPackages.hasTypingsFor` method. It felt like getting a `false` response shouldn’t push errors onto the `allPackages.errors` there—you should be allowed to ask for typings that don’t exist and deal with the `undefined` right away. On the other hand, if you do `await allPackages.allTypings()` and some had errors, you may not want it to throw immediately—the point of this PR is to be able to ignore a lot of stuff that’s not relevant to a partial run in CI. So those errors are stored and checked by `checkParseResults`.

packages/definitions-parser/src/packages.ts
andrewbranch: I initially had `errors` private... I think I used it externally in some validation that ended up being redundant and I removed it. Will take a look.

packages/definitions-parser/src/packages.ts
andrewbranch: > what kind of errors actually get thrown here

This is doing way more than reading. This is the process where we create TS source files and perform module resolution on imports/references so we can see what files are used/unused and check that packages aren’t using relative imports of other packages.

packages/publisher/package.json
jakebailey: Sent #767 since it took 2 seconds.

jakebailey (approved):
At this point, nothing in particular stands out to me. But Nathan's definitely more familiar.

Probably I should release https://github.com/microsoft/DefinitelyTyped-tools/pull/766 first just to have a nice working build?

packages/dtslint-runner/src/prepareAllPackages.ts
```diff
@@ -1,29 +1,27 @@
...
     await installAllDependencies(definitelyTypedPath);
   }
-  const errors = checkParseResults(allPackages);
+  const errors = await checkParseResults(allPackages);
   if (errors.length) {
     throw new Error(errors.join("\n"));
   }
   return {
-    packageNames: new Set(allPackages.allTypings().map(({ subDirectoryPath }) => subDirectoryPath)),
+    packageNames: new Set((await allPackages.allTypings()).map(({ subDirectoryPath }) => subDirectoryPath)),
```
sandersn, line 24: I *kind* of like how `async` has become a proxy for "performs IO" in this code.

packages/definitions-parser/src/lib/definition-parser.ts
sandersn: Yeah, I think my later suggestion to have two string types is the right approach actually.

sandersn (approved):

PR #765, opened 2023-10-19
Add changeset for #757

jakebailey:
#757 was missing changesets, so got partially published breaking downstream importers:

```
node_modules/.pnpm/@definitelytyped+definitions-parser@0.0.180_typescript@5.3.0-dev.20231018/node_modules/@definitelytyped/definitions-parser/dist/packages.d.ts:1:10 - error TS2305: Module '"@definitelytyped/header-parser"' has no exported member 'Contributor'.

1 import { Contributor, Header, License } from "@definitelytyped/header-parser";
           ~~~~~~~~~~~
```

### DIFF
--- a/.changeset/thick-tigers-cry.md
+++ b/.changeset/thick-tigers-cry.md
@@ -1,0 +1,7 @@
+---
+"@definitelytyped/definitions-parser": patch
+"@definitelytyped/header-parser": patch
+"@definitelytyped/publisher": patch
+---
+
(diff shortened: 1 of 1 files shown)


andrewbranch (approved):

jakebailey:
Note that this is technically wrongish because I'm bumping packages that would have already been bumped for the same change but indirectly, but doing all of the above is a good idea IMO.

andrewbranch:
Don’t cry, thick tigers 😢 who did this to you

PR #767, opened 2023-10-19
Drop use of source-map-support

jakebailey:
See https://github.com/microsoft/DefinitelyTyped-tools/pull/764#discussion_r1365903059; Node has had this support in a non-experimental state since Node 14.18.


andrewbranch (approved):

pnpm-lock.yaml
```diff
@@ -343,12 +340,9 @@ importers:
       semver:
         specifier: ^7.3.7
         version: 7.5.2
-      source-map-support:
-        specifier: ^0.4.0
-        version: 0.4.18
       typescript:
         specifier: next
-        version: 5.3.0-dev.20231018
+        version: 5.3.0-dev.20231019
```
jakebailey, line 345: Unrelated changes from pnpm re-resolving the tree on changes. Unfortunate now that we are using `@next`.

PR #771, opened 2023-10-20
Lazy `AllPackages` (2)

andrewbranch:
#764 + fixes in the last couple commits


packages/publisher/src/generate-packages.ts
```diff
@@ -36,7 +36,7 @@ if (require.main === module) {
   const tgz = !!yargs.argv.tgz;
   logUncaughtErrors(async () => {
     const log = loggerWithErrors()[0];
-    const options = { ...defaultLocalOptions, definitelyTypedPath: outputDirPath };
+    const options = { ...defaultLocalOptions };
```
jakebailey, line 41: Hm, how did this work before? I noticed this before when debugging publishing the first time and wasn't sure why we were setting this to an output path...

andrewbranch: No idea! I don’t think I’ve ever run `pnpm generate` before but it was definitely broken for me

jakebailey (approved):

packages/definitions-parser/src/lib/definition-parser.ts
```diff
@@ -254,7 +254,7 @@ export function getFiles(
   moduleResolutionHost: ts.ModuleResolutionHost
 ): readonly FilesForSingleTypeScriptVersion[] {
   const errors = [];
-  const rootDir = dt.subDir("types").subDir(typingsData.typesDirectoryName);
+  const rootDir = dt.subDir("types").subDir(typingsData.subDirectoryPath);
```
sandersn, line 257: this is the version that includes /v15, eg `react/v15`, right? `rootDir` sounds like a misleading name for it.

sandersn (approved):

andrewbranch:
I’ll let you guys decide if/when to merge this, as I may not have the chance to watch it closely today. I tested generating packages locally so I think it’s good, but maybe you’d prefer to merge it Monday 🤷

sandersn:
I'd prefer Monday, or Wednesday when both Jake and I will have a chance to watch it.

jakebailey:
Fixed merge conflicts and also removed the references to `parse` from the CI jobs, which weren't updated. This also fixes #780 so probably want to try and merge tomorrow morning.

sandersn (approved):

PR #773, opened 2023-10-20
node16 test fallbacks for old TS versions, fix dependency/devDependency import

### LINKED ISSUE #772: Allow imports of deps listed both in devDependencies and in dependencies
I was trying to lock down the version of `marked` for testing `marked-terminal` by specifying a specific version in `devDependencies` while `dependencies` had a large range:

```
    "dependencies": {
        "@types/cardinal": "^2.1",
        "@types/node": "*",
        "chalk": "^5.3.0",
        "marked": ">=6.0.0 <10"
    },
    "devDependencies": {
        "@types/marked-terminal": "workspace:.",
        "marked": "6.0.0"
    },
```

but @definitelytyped/no-import-of-dev-dependencies complains about imports of `marked` after this.

andrewbranch:
Fixes #772 and https://github.com/DefinitelyTyped/DefinitelyTyped/pull/67090#discussion_r1367525696

### DIFF
--- a/packages/eslint-plugin/src/rules/no-import-of-dev-dependencies.ts
+++ b/packages/eslint-plugin/src/rules/no-import-of-dev-dependencies.ts
@@ -26,8 +26,11 @@ const rule = createRule({
       const devdeps = packageJson
         ? Object.keys(packageJson.devDependencies).map((dep) => dep.replace(/@types\//, ""))
         : [];
+      const deps = packageJson
+        ? Object.keys(packageJson.dependencies ?? {}).map((dep) => dep.replace(/@types\//, ""))
+        : [];
       commentsMatching(context.getSourceCode(), /<reference\s+types\s*=\s*"(.+)"\s*\/>/, (ref, comment) => {
-        if (devdeps.includes(ref) && ref !== packageName) {
+        if (devdeps.includes(ref) && ref !== packageName && !deps.includes(ref)) {
           report(comment, "noReferenceOfDevDependencies");
         }
       });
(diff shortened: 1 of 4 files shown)


andrewbranch:
I looked at the tests for the eslint rule and had no idea how to add to it because I didn’t see any package.json content.

packages/eslint-plugin/src/rules/no-import-of-dev-dependencies.ts
```diff
@@ -26,16 +26,19 @@ const rule = createRule({
       const devdeps = packageJson
         ? Object.keys(packageJson.devDependencies).map((dep) => dep.replace(/@types\//, ""))
         : [];
+      const deps = packageJson
+        ? Object.keys(packageJson.dependencies ?? {}).map((dep) => dep.replace(/@types\//, ""))
+        : [];
```
jakebailey, line 31: Preexisting but isn't this wrong when the package is scoped?

andrewbranch: Yep

jakebailey: 😬

sandersn:
> I looked at the tests for the eslint rule and had no idea how to add to it because I didn’t see any package.json content. Tested both of these by linking locally. Feel free to push tests while I’m gone if you want.

My memory is that I couldn't figure out how to test these either; another reason to figure out how to switch to an existing rule.

jakebailey (approved):
LGTM but needs changesets for the two changed packages.

jakebailey:
> I looked at the tests for the eslint rule and had no idea how to add to it because I didn’t see any package.json content.

Oh, and on this is because we misuse the package.json for the plugin itself as a source of tests... Very silly, and my other PR also helps that too (if we port every test to a fixture).

jakebailey (approved):

PR #776, opened 2023-10-23
Fix publish of interdependent packages

jakebailey:
If we're publishing types packages which depend on each other, `pacote` isn't going to resolve to them. I'm not sure how this worked before (this code appears to have existed before the pnpm branch); for now this unblocks publishing by not checking types packages which exist in the repo (albeit without a version check).


sandersn:
It's a bug in the new code, but we didn't try to publish a package with *new* inter-DT deps in the last week, so bbob is the first thing to hit it. Your fix is the right one; check allPackages instead of npm, since we assert elsewhere that the state of the DT repo itself is good.

The reason it worked all week is that a previously-published `@types/X` is already on npm.
The reason it worked pre-pnpm is that `@types` deps were only in `dependencies` if they were out of date ... and were already on npm.

PR #777, opened 2023-10-23
Disable corepack when we run npm

jakebailey:
This would unblock https://github.com/DefinitelyTyped/DefinitelyTyped/pull/67120, but I actually think these two uses could just be `pacote` + https://www.npmjs.com/package/@andrewbranch/untar.js or something.

### DIFF
--- a/.changeset/mighty-falcons-matter.md
+++ b/.changeset/mighty-falcons-matter.md
@@ -1,0 +1,5 @@
+---
+"@definitelytyped/dts-critic": patch
+---
+
+Disable corepack when we run npm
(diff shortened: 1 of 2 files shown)


sandersn (approved):

sandersn:
I don't know how we can use untar.js since it's not labelled Blazing Fast.

jakebailey:
We wouldn't actually need the tar lib as pacote can extract. But this code unfortunately has to be synchronous because it's called from tslint anyhow

jakebailey:
Forgot the changeset, oops

PR #779, opened 2023-10-25
pnpm install retag workflow

sandersn:
update-ts-version-tags.yml

### DIFF
--- a/.github/workflows/update-ts-version-tags.yml
+++ b/.github/workflows/update-ts-version-tags.yml
@@ -34,9 +34,9 @@ jobs:
     - name: retag
       run: |
         git clone --depth 1 https://github.com/DefinitelyTyped/DefinitelyTyped ../DefinitelyTyped
-      - run: pnpm install
-        working-directory: ../DefinitelyTyped
-      - run: node packages/retag/dist/index.js
+    - run: pnpm install
+      working-directory: ../DefinitelyTyped
+    - run: node packages/retag/dist/index.js
       env:
         NPM_TOKEN: ${{ secrets.NPM_RETAG_TOKEN }}
         GH_API_TOKEN: ${{ secrets.GH_API_TOKEN }}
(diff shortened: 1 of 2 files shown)


.github/workflows/update-ts-version-tags.yml
```diff
@@ -31,10 +31,13 @@ jobs:
     - run: pnpm config set store-dir $PNPM_CACHE_FOLD
```
sandersn, line 31: looking at the preceding environment variables, this looks like it's missing a trailing `ER` to make it $PNPM_CACHE_FOLDER. @jakebailey is that correct?

jakebailey: This (and other actions) should actually be updated to use the caching mechanism I added in #762.

.github/workflows/update-ts-version-tags.yml
```diff
@@ -31,10 +31,13 @@ jobs:
...
     - run: pnpm install
     - run: pnpm build
-    - name: retag
+    - name: clone
       run: |
         git clone --depth 1 https://github.com/DefinitelyTyped/DefinitelyTyped ../DefinitelyTyped
```
jakebailey, line 36: This should probably use the checkout action, though I'm not entirely certain that supports relative paths. IMO we should be trying (long term) to not depend on DT being in the parent dir of the DT-tools codebase (at least, it's a little too much cognitive overhead for me compared to passing `--path`).

sandersn: - Is there another workflow that does this? I'll go look at the other ones again.
- Would putting in `DefinitelyTyped` instead of `../DefinitelyTyped` be OK? I can easily add a `--path` option to retag.

sandersn: ok, copied the checkout code from publish-packages.yml

jakebailey: It's fine to not change the cloning, no big deal, that took a code change in the publisher which I did because I just didn't remember / trust myself to check things out in the right place...

jakebailey (approved):

PR #786, opened 2023-10-26
Fix dependent packages showing up as changed in getAffectedPackages

jakebailey:
Fixes https://github.com/DefinitelyTyped/DefinitelyTyped/issues/67205

I think I probably incorrectly suggested the bad code in a change, not understanding what this code did. This just removes `...` which gives the right answer on DT:

```
dtslint-runner@0.0.188
Node version:  v20.9.0
Using local DefinitelyTyped at .
Running: git rev-parse --verify master
d0df83268470773861ff1bd2a6a9520b8c247d3d
Running: git diff master --name-status
M       types/node/v18/child_process.d.ts
M       types/node/v18/test/child_process.ts
Testing 1 changed packages: Set(1) { 'node/v18' }
Testing 9 dependent packages: Set(9) {
  'ssh2',
  'docker-modem',
  'dockerode',
  'docker-events',
  'node-ssh',
  'promise-sftp',
  'ssh2-sftp-client',
  'ssh2-sftp-client/v7',
  'tunnel-ssh'
}
```

### DIFF
--- a/packages/definitions-parser/src/get-affected-packages.ts
+++ b/packages/definitions-parser/src/get-affected-packages.ts
@@ -56,7 +56,6 @@ export async function getAffectedPackages(
       )
     );
   }
-
   return getAffectedPackagesWorker(
     allPackages,
     changedPackageDirectories,
--- a/packages/definitions-parser/src/get-affected-packages.ts
+++ b/packages/definitions-parser/src/get-affected-packages.ts
@@ -16,8 +16,9 @@ export async function getAffectedPackages(
   definitelyTypedPath: string
 ): Promise<{ errors: string[] } | PreparePackagesResult> {
   const errors = [];
+  // No ... prefix; we only want packages that were actually edited.
   const changedPackageDirectories = await execAndThrowErrors(
-    `pnpm ls -r --depth -1 --parseable --filter '...@types/**[${sourceRemote}/${sourceBranch}]'`,
+    `pnpm ls -r --depth -1 --parseable --filter '@types/**[${sourceRemote}/${sourceBranch}]'`,
     definitelyTypedPath
   );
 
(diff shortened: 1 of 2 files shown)


packages/definitions-parser/src/get-affected-packages.ts
```diff
@@ -52,6 +56,11 @@ export async function getAffectedPackages(
       )
     );
   }
+
+  console.log("changedPackageDirectories", changedPackageDirectories);
+  console.log("addedPackageDirectories", addedPackageDirectories);
+  console.log("allDependentDirectories", allDependentDirectories);
```
sandersn, line 62: Don't we log similar derived information a little bit later? It's useful but probably overwhelming especially for large changes.

No recommendation, just an observation. More logging is better than none, so I guess leave it in.

sandersn (approved):

packages/definitions-parser/src/get-affected-packages.ts
jakebailey: forgot to push, this is not supposed to be there!

packages/definitions-parser/src/get-affected-packages.ts
```diff
@@ -52,6 +56,7 @@ export async function getAffectedPackages(
       )
     );
   }
+
```
jakebailey, line 59:
```suggestion
```

packages/definitions-parser/src/get-affected-packages.ts
jakebailey: (pushed now...)

jakebailey:
Meta point, but would be nice to set up some sort of test DT that we can test these pnpm commands in or something.

peterblazejewicz:
thanks!

PR #789, opened 2023-10-27
Throw parse errors early now that more stuff is lazy

jakebailey:
See: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/67218/files#r1374846390

If things break in the parse stage, we may completely crash later. Now that parsing skips a bunch of work like file loading, I think that bailing early for total parse failures is okay.

While here, add `debugger` to the assert helper, and fix an issue where `execAndThrowErrors` was including stderr, which breaks debug attaching because `stderr` contains "waiting for debugger to exit", breaking command result parsing.


andrewbranch (approved):
The value of waiting to throw errors is so you can see all the reasons CI was going to fail, not just the first one, and fix them in fewer iterations. But if that’s not particularly relevant here, this is fine.

jakebailey:
Largely the problem is that a failure to load may cause other failures later because packages which fail to load go into some `errors` list and then throw when you try to touch them.

andrewbranch:
My intent was to make it so that if someone accidentally checks in something really broken, work on completely unrelated packages doesn't have to grind to a halt.

jakebailey:
Yeah, that makes sense, I'm just not sure how to balance that along with not crashing later when a package that is touched has an error that we haven't exposed yet...

jakebailey:
FWIW I'm hoping we can make as much of this whole loading process as infallible as possible, especially now that file detection is going away. At that point we're just a fancy `package.json` parser.

Issue #734, opened 2023-09-15
no-unnecessary-generics reporting in a less useful place than it could?


johnnyreilly:
Hi all!

I've been looking into this [Definitely Typed pull request](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66531) and picking through the issues with @JoshuaKGoldberg.  We encountered an interesting case with the `no-unnecessary-generics` rule which we think could be better.  Consider:

```ts
    <TElement extends HTMLElement = HTMLElement>(
      html: JQuery.htmlString,
      ownerDocument_attributes?: Document | JQuery.PlainObject,
      // eslint-disable-next-line @definitelytyped/no-unnecessary-generics
    ): JQuery<TElement>;
```

You'll note that we have an `eslint-disable-next-line` just before the last line; the ReturnType of the method.  Without this in place the red squigglies would be under:

```ts
JQuery<TElement>
       ~~~~~~~~
```

This isn't the useful place to have this report.  Perhaps the better place would be the first mention of `TElement`, like so:

```ts
    <TElement extends HTMLElement = HTMLElement>(
     ~~~~~~~~
```

Looking at https://github.com/microsoft/DefinitelyTyped-tools/blob/822dadbed4d2946f399400fd512599d0a2505d51/packages/eslint-plugin/src/rules/no-unnecessary-generics.ts#L57 it seems this is intentional.  

Is this good? What do you think?

JoshuaKGoldberg:
FWIW I'd lean towards reporting on the type parameter itself. This behavior confused & surprised me at first.

PR #794, opened 2023-10-30
fix: position of reporting ESLint rule no-unnecessary-generics

johnnyreilly:
This PR addresses https://github.com/microsoft/DefinitelyTyped-tools/issues/734 and is a result of pairing with the ever kind and generous @joshuakgoldberg.

When working on the formatting of Definitely Typed, we happened upon some surprising behaviour around the `no-unnecessary-generics` rule. In the words of George Bernard Shaw: "I often quote myself. It adds spice to my conversation."  Let me quote me now:

> You'll note that we have an `eslint-disable-next-line` just before the last line; the ReturnType of the method.  Without this in place the red squigglies would be under:
> 
> ```ts
> JQuery<TElement>
>        ~~~~~~~~
> ```
> 
> This isn't the useful place to have this report.  Perhaps the better place would be the first mention of `TElement`, like so:
> 
> ```ts
>     <TElement extends HTMLElement = HTMLElement>(
>      ~~~~~~~~
> ```

This PR implements the suggested change; moving the position of the report to the type parameter position.

What do you think?

### DIFF
--- a/packages/eslint-plugin/src/rules/no-unnecessary-generics.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-generics.ts
@@ -47,16 +47,15 @@ const rule = createRule({
         const checker = parserServices.program.getTypeChecker();
 
         for (const typeParameter of tsNode.typeParameters) {
-          const name = typeParameter.name.text;
-          const res = getSoleUse(tsNode, assertDefined(checker.getSymbolAtLocation(typeParameter.name)), checker);
-          
-          if (res.type === "ok") {
+          const result = getSoleUse(tsNode, assertDefined(checker.getSymbolAtLocation(typeParameter.name)), checker);
+
+          if (result === "ok") {
             continue;
           }
 
           context.report({
-            data: { name },
-            messageId: res.type,
+            data: { name: typeParameter.name.text },
+            messageId: result,
             node: parserServices.tsNodeToESTreeNodeMap.get(typeParameter),
           });
         }
@@ -65,7 +64,8 @@ const rule = createRule({
   },
 });
 
-interface Result { type: "ok" | "never" | "sole"; }
+type Result = "ok" | "never" | "sole";
+
 function getSoleUse(sig: ts.SignatureDeclaration, typeParameterSymbol: ts.Symbol, checker: ts.TypeChecker): Result {
   const exit = {};
   let soleUse: ts.Identifier | undefined;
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin/src/rules/no-unnecessary-generics.ts
```diff
@@ -54,7 +54,7 @@ const rule = createRule({
               context.report({
                 data: { name },
                 messageId: "sole",
-                node: parserServices.tsNodeToESTreeNodeMap.get(res.soleUse),
```
jakebailey, line 57: I think you should be able to drop `soleUse` or simplify otherwise, actually; I think it's only returned for this particular case.

jakebailey:
I am personally all for this, though I am guessing it errored on the use just so that you could easily see where the single use was. But, I don't personally think that's too helpful as usually the fix is to remove the generic, thereby making the one use error anyway.

packages/eslint-plugin/src/rules/no-unnecessary-generics.ts
johnnyreilly: Lemme have a looksee - just grabbing some dinner

johnnyreilly: simplified - think this is what you were suggesting?

jakebailey: Yes, even further than I was suggesting, even! (Which is not bad.)

JoshuaKGoldberg: Let's go even further 🔪 no need for an interface with just the one property. `type Result = "ok" | ...`!

JoshuaKGoldberg: ...or even `MessageId | undefined` 🔪 🔪

johnnyreilly: gosh you guys are brutal.  Let me sharpen the knife

johnnyreilly: how's this?

JoshuaKGoldberg: Looks great!

> `feat: 🔪`

nice touch :)

jakebailey:
FWIW you should also write a changeset for this PR (`pnpm changeset`, and just let it do a patch bump). We don't have the bot set up (because it would tell everyone modifying the dependency allowlist to add changelist which we don't want); sorry about that.

johnnyreilly:
Added the changeset - hope that's right?

jakebailey:
Seems correct to me.

JoshuaKGoldberg (approved):
(in case this is helpful at all)

sandersn (approved):

PR #799, opened 2023-10-31
Enforce changesets in PRs (excluding allowedPackageJsonDependencies)

jakebailey:
This adds a CI check that verifies that PRs contain changesets if they modify packages, ignoring PRs which only modify `allowedPackageJsonDependencies.txt`.

```
🦋  error Some packages have been changed but no changesets were found. Run `changeset add` to resolve this error.
🦋  error If this change doesn't need a release, run `changeset add --empty`.
```

### DIFF
--- a/packages/definitions-parser/allowedPackageJsonDependencies.txt
+++ b/packages/definitions-parser/allowedPackageJsonDependencies.txt
@@ -3,7 +3,6 @@
 @azure/functions
 @babel/code-frame
 @babel/core
-@babel/core
 @babel/generator
 @babel/parser
 @babel/template
--- a/packages/definitions-parser/allowedPackageJsonDependencies.txt
+++ b/packages/definitions-parser/allowedPackageJsonDependencies.txt
@@ -3,6 +3,7 @@
 @azure/functions
 @babel/code-frame
 @babel/core
+@babel/core
 @babel/generator
 @babel/parser
 @babel/template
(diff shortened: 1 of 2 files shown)


packages/definitions-parser/allowedPackageJsonDependencies.txt
```diff
@@ -3,6 +3,7 @@
 @azure/functions
 @babel/code-frame
 @babel/core
+@babel/core
```
jakebailey, line 6:
```suggestion
```

oops

andrewbranch (approved):
Nice workaround! Is publisher already excluded by merit of being `"private": true`?

jakebailey:
No, it isn't excluded and gets full changelogs and tagging:

- https://github.com/microsoft/DefinitelyTyped-tools/blob/master/packages/publisher/CHANGELOG.md
- https://github.com/microsoft/DefinitelyTyped-tools/releases/tag/%40definitelytyped%2Fpublisher%400.0.185

To exclude it, I could exclude it in the changeset config (and probably delete its changelog and version number). https://github.com/changesets/changesets/issues/1093

jakebailey:
Confusingly, `retag` is _not_ private. That is probably an oops from a while ago...

andrewbranch:
I feel like publisher (and retag) don’t need to be version managed, but I'm fine either way

jakebailey:
I agree. I'll send a different PR to fix that up.

jakebailey:
I didn't read the linked thread in full; apparently ignore doesn't achieve what I hoped. But, I can at least make retag private.

PR #802, opened 2023-10-31
Format missed files

jakebailey:
Not sure if we should just update the auto format action to format all files; wasn't sure if that would be dangerous or not.


sandersn (approved):

sandersn (approved):
What even is .knip.json and why is it hidden.

jakebailey:
https://github.com/microsoft/DefinitelyTyped-tools/pull/792; a cool tool to check for dead code and other stuff.

PR #809, opened 2023-11-03
Add ts 5.4

sandersn:


jakebailey (approved):

sandersn:
A lot fewer tests than before. Is that because we're doing less work to transform `typesVersions` now, and just check up-front that they're correct?

PR #812, opened 2023-11-04
[definitions-parser] skip scripts folder

KhooHaoYit:
Partially unblocks DefinitelyTyped/DefinitelyTyped#67259 and DefinitelyTyped/DefinitelyTyped#67291

When doing type checking on CI, the package that depends on it might use it in the `scripts` folder while not using it in the `types` folder, causing the CI to fail from missing dependencies due to `pnpm` only installs packages which got affected

In DefinitelyTyped/DefinitelyTyped#67259 case `pnpm` installed packages for
```
es-abstract/scripts
get-intrinsic/scripts
lodash-es/scripts
material-ui/scripts
...
```
While the CI checks typing for
```
es-abstract
get-intrinsic
lodash-es
material-ui
...
```
This PR fixes that issue by skipping type checking for `scripts` folder

### DIFF
--- a/.changeset/five-bears-tickle.md
+++ b/.changeset/five-bears-tickle.md
@@ -1,0 +1,5 @@
+---
+"@definitelytyped/definitions-parser": patch
+---
+
+Don't treat script dirs as types packages
--- a/packages/definitions-parser/src/packages.ts
+++ b/packages/definitions-parser/src/packages.ts
@@ -614,20 +614,30 @@ export function getDependencyFromFile(
     return undefined;
   }
 
-  const [typesDirName, name, subDirName] = parts; // Ignore any other parts
+  // types/[packageName]/[scripts]
+  // types/[packageName]/[tsVersion]/[scripts]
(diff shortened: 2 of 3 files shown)


KhooHaoYit:

jakebailey:
FWIW I would really like to stop (generally) using path only information in the repo and instead determine paths are part of which package by looking for and reading package.json files (something I have done in #788. I don't really want to special case script does in specific directories when we know exactly which folders are packages based on their proximity to a package.json with the right name.

This might be okay as a hack, but it's not wrong for a package to contain a folder named "scripts" that's actually meant to be packaged.

jakebailey:
(to be clear it may be okay for a hack in the short term, but I'm just noting my thought before I forget)

jakebailey:
This PR will need a changeset to be accepted. I'd push to your PR to add one, but you've sent the PR from your master branch which is a little wonky, but I can still push.

jakebailey (approved):

PR #815, opened 2023-11-06
Don't clone or install in CI

jakebailey:
Now that files are static, we no longer need to do this.

### DIFF
--- a/.github/workflows/publish-packages.yml
+++ b/.github/workflows/publish-packages.yml
@@ -41,14 +41,7 @@ jobs:
           key: pacote-cache-${{ github.run_id }}-${{ github.run_attempt }}
           restore-keys: pacote-cache-
 
-      - uses: actions/checkout@v4
-        with:
-          repository: 'DefinitelyTyped/DefinitelyTyped'
-          path: 'DefinitelyTyped'
-      - run: pnpm install
-        working-directory: DefinitelyTyped
-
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-packages.yml
```diff
@@ -41,14 +41,7 @@ jobs:
           key: pacote-cache-${{ github.run_id }}-${{ github.run_attempt }}
           restore-keys: pacote-cache-
 
-      - uses: actions/checkout@v4
-        with:
-          repository: 'DefinitelyTyped/DefinitelyTyped'
-          path: 'DefinitelyTyped'
```
sandersn, line 47: why don't we have to clone? Don't the files have to be on disk somewhere to publish them?

sandersn (changes requested):

.github/workflows/publish-packages.yml
jakebailey: Nope; we didn't have to clone before pnpm. When we switched, things broke because we could no longer analyze dependencies through node_modules that didn't exist. Now the files are static and we don't need to use TS anymore so this is legal again.

sandersn (approved):

.github/workflows/publish-packages.yml
andrewbranch: Instead of cloning, `getDefinitelyTyped` streams the DT tarball from GitHub into memory.

andrewbranch (approved):

PR #818, opened 2023-11-07
Update / clean up various dependencies

jakebailey:
Not quite sure how to best send something like this, but I've been slowly updating dependencies, removing `fs-extra`, etc.


jakebailey:
I undrafted this for review, but it's currently in a bad state because our tests are not actually within any `tsconfigs` so newer ts-eslint fails as it cannot be imported using `node10` resolution. Need to get these covered somehow but not quite sure.

andrewbranch:
All of the packages I migrated to DT-tools when I first set up the monorepo have a test-specific tsconfig.json. If other packages are missing a tsconfig, we should adopt that pattern. Then we can start incrementally migrating to nodenext. (Or all in one go if it’s easy, but I was anticipating some broken dependencies.)

jakebailey:
I'll take a look to see what you did there; the main problem right now are the lint rule tests. Funkily that one has a tsconfig but it is for the test cases themselves. Probably what I should do instead is work on the lint rule baselines so we don't need those tests in their current form.

andrewbranch (approved):

jakebailey:
Will probably wait until Monday to do this; this upgrades our version of eslint and ts-eslint which should theoretically be fine but better safe than sorry on a Friday.

jakebailey:
After this PR:

```console
$ pnpm outdated -r             
┌───────────────────┬──────────┬────────────┬────────────────────────────────┐
│ Package           │ Current  │ Latest     │ Dependents                     │
├───────────────────┼──────────┼────────────┼────────────────────────────────┤
│ tslint            │ 6.1.3    │ Deprecated │ @definitelytyped/dtslint       │
├───────────────────┼──────────┼────────────┼────────────────────────────────┤
│ @octokit/rest     │ 19.0.13  │ 20.0.2     │ @definitelytyped/dtslint-      │
│                   │          │            │ runner                         │
├───────────────────┼──────────┼────────────┼────────────────────────────────┤
│ @types/node       │ 16.18.61 │ 20.9.0     │ @definitelytyped/definitions-  │
│                   │          │            │ parser, @definitelytyped/utils │
├───────────────────┼──────────┼────────────┼────────────────────────────────┤
│ @types/node (dev) │ 16.18.61 │ 20.9.0     │ @definitelytyped/retag,        │
│                   │          │            │ definitelytyped-tools          │
└───────────────────┴──────────┴────────────┴────────────────────────────────┘
```

Obviously we know about tslint. Node is held back since we're only targeting Node 16, though IMO it's fine to go to 20. `@octokit/rest` is set back to the oldest version used in CI but we could bump that to a later version if we raise DT-tools' minimum to Node 18. But, we only use that script for the script that posts the DT diffs to the TS repo on demand.

jakebailey:
Running DT against this... guess some new rules are enabled and things fail. will fix

PR #824, opened 2023-11-14
Run real ESLint in test, baseline output

jakebailey:
This moves all of our eslint tests into the "fixtures" folder. In the test, it runs eslint on each file, baselining the output in a human-readable format with error ranges and so on (like the TS error deltas).

This exposed some incorrect error ranges (and probably more; I need to skim the changes again).


packages/eslint-plugin/test/__file_snapshots__/types/no-const-enum/index.d.ts.lint
```diff
@@ -0,0 +1,5 @@
+No errors
+
+==== types/no-const-enum/index.d.ts ====
+
+    enum F {}
```
jakebailey, line 5: I snapshotted even the "good" files; the runner doesn't do a great job of removing unused snapshots (at least in this special "file snapshot" addon).

package.json
```diff
@@ -40,6 +40,9 @@
         "@types/strip-json-comments",
         "strip-json-comments"
       ]
+    },
+    "patchedDependencies": {
+      "jest-file-snapshot@0.5.0": "patches/jest-file-snapshot@0.5.0.patch"
```
jakebailey, line 45: This is due to https://github.com/satya164/jest-file-snapshot/pull/23 (which I only found after I had patched it locally myself).

andrewbranch (approved):
This is great, thanks!

JoshuaKGoldberg:
cc @bradzacher, this is beautiful 🥲

packages/eslint-plugin/test/__file_snapshots__/types/no-const-enum/index.d.ts.lint
jakebailey: Unused baselines are now caught via a test, but I think leaving these to make sure we encode the expected state is probably valuable.

bradzacher:
Yeah it's hard because doing this in the general case is much harder than a per-repo case. People have certain expectations about automation and on top of that the rule tester is framework-agnostic.

One-file-per-case is a really really nice way to do testing of zero-option rules like we've got here - but when you need to test options it quickly balloons out into an unmanageable mess of copy-pasted files that are hard to grep through to find duplicates and such.

I still want to solve this from the framework side so that we can provide such tooling to everyone - we just need to do some thinking and design work.

jakebailey:
> One-file-per-case is a really really nice way to do testing of zero-option rules like we've got here - but when you need to test options it quickly balloons out into an unmanageable mess of copy-pasted files that are hard to grep through to find duplicates and such.

In the case of this test setup, the fixtures folder is a valid eslint root such that we can just make more directories and modify things via eslintrcs in each. It also means I can open an editor to the side in the directory and see live what's going on (which was handy while fixing the off by ones...)

bradzacher:
Yeah sorry I meant as an example [this rule's tests](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/tests/rules/member-delimiter-style.test.ts).

The `member-delimiter-style` rule has over 150 test cases - many of which are duplicated code just with different configurations. In this case if you need to setup one file and one eslint config per test - that's over 300 files right there!

Now finding duplicates you have to do cross-file searches as opposed to same-file searches (there's a [lint rule](https://github.com/eslint-community/eslint-plugin-eslint-plugin/blob/main/docs/rules/no-identical-tests.md) which reports on duplicate test cases). Changing duplicate cases becomes a multi-file find-and-replace as opposed to a multi-cursor edit.

That's not even our biggest rule test suite either!

That's what I meant by hard to manage - when you have to manage options and re-testing the same code under different variations things can be a little tricky as the space scales exponentially!
Compared to rules in this plugin which which have no options because they're designed for one usecase.

I'm not saying this style is bad at all btw - I like it a lot! We use this style as well for our [ast fixtures](https://github.com/typescript-eslint/typescript-eslint/tree/main/packages/ast-spec/src/declaration/ClassDeclaration/fixtures).
The thing I'm saying is that as a "the general case solution for lint rule testing" it doesn't work super well. Trying to solve the more general case is *hard*.

PR #830, opened 2023-11-20
Port builtin tslint rules

sandersn:
Convert the "good" tslint builtins to their equivalent eslint or typescript-eslint equivalents.
I tried to use no-redeclare to prevent duplicated `var` declarations, but the eslint rule doesn't understand Typescript syntax and ts-eslint rule doesn't understand Typescript semantics that well (and, a little, its syntax), so had too many false positives.

Obviously there are other rules that would be nice to enable, but I'll leave that for others, since there is a lot of testing effort per rule, and these 10 are the ones @jakebailey, @andrewbranch and I thought were the most useful.


packages/eslint-plugin/src/configs/all.ts
```diff
@@ -146,7 +146,34 @@ export const all: Linter.BaseConfig = {
...
+        "@typescript-eslint/no-empty-interface": "error",
+        "no-duplicate-imports": "error",
+        "@typescript-eslint/array-type": ["error", { default: "array-simple" }],
+        "@typescript-eslint/naming-convention": [
+          "error",
+          {
+            selector: "interface",
+            format: [],
+            custom: {
+              regex: "^I[A-Z]",
```
jakebailey, line 169: Which rule provided this before? Is it definitely this regex? This seems kinda strict but maybe we already pass with it.

packages/eslint-plugin/test/__file_snapshots__/types/no-import-of-dev-dependencies/bad.d.ts.lint
```diff
@@ -1,17 +1,23 @@
...
+  5:1   error  .d.ts files may not import packages in devDependencies                           @definitelytyped/no-import-of-dev-dependencies
 
-✖ 4 problems (4 errors, 0 warnings)
+✖ 6 problems (6 errors, 0 warnings)
 
 ==== types/no-import-of-dev-dependencies/bad.d.ts ====
 
     /// <reference types="devdep"/>
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+!!! @typescript-eslint/triple-slash-reference: Do not use a triple slash reference for devdep, use `import` style instead.
```
jakebailey, line 15: I assume that this only appears because I also import that same name from this file below, and the rule is doing what we want otherwise?

jakebailey (approved):

packages/eslint-plugin/src/configs/all.ts
sandersn: It's directly based on the recommended port from `interface-name-prefix` documented on the ts-eslint site:
https://typescript-eslint.io/rules/naming-convention#enforce-that-interface-names-do-not-begin-with-an-i

packages/eslint-plugin/test/__file_snapshots__/types/no-import-of-dev-dependencies/bad.d.ts.lint
sandersn: Yeah, the rule is working as expected, and I thought it was better to show the error than to add a suppression comment. Let me know if you want it the other way round, though.

jakebailey: Nah, I'd leave it. Nice to have the baseline IMO

PR #835, opened 2023-11-21
Fix getAffectedPackages for PRs which modify only versioned dirs

jakebailey:
Unblocks https://github.com/DefinitelyTyped/DefinitelyTyped/pull/67463

We really need some sort of test suite for this code.

Also, I really am surprised that we don't have this helper somewhere else. That and I'm eternally confused by `typesDirectoryName` versus `subDirectoryPath`, as well as packages which are missing dirs and versions?

### DIFF
--- a/packages/definitions-parser/src/get-affected-packages.ts
+++ b/packages/definitions-parser/src/get-affected-packages.ts
@@ -29,7 +29,11 @@ export async function getAffectedPackages(
     return { errors };
   }
   const { additions, deletions } = git;
-  const addedPackageDirectories = mapDefined(additions, (id) => id.typesDirectoryName);
+  const addedPackageDirectories = mapDefined(additions, (pkg) => {
+    if (!pkg.typesDirectoryName) return undefined;
+    if (!pkg.version || pkg.version === "*") return pkg.typesDirectoryName;
+    return `${pkg.typesDirectoryName}/v${formatTypingVersion(pkg.version)}`;
+  });
   const allDependentDirectories = [];
   // Start the filter off with all packages that were touched along with those that depend on them.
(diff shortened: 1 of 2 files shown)


packages/definitions-parser/src/get-affected-packages.ts
```diff
@@ -29,7 +29,11 @@ export async function getAffectedPackages(
     return { errors };
   }
   const { additions, deletions } = git;
-  const addedPackageDirectories = mapDefined(additions, (id) => id.typesDirectoryName);
+  const addedPackageDirectories = mapDefined(additions, (pkg) => {
+    if (!pkg.typesDirectoryName) return undefined;
+    if (!pkg.version || pkg.version === "*") return pkg.typesDirectoryName;
+    return `${pkg.typesDirectoryName}/v${formatTypingVersion(pkg.version)}`;
+  });
```
andrewbranch, line 36:
```suggestion
  const addedPackageDirectories = mapDefined(additions, (pkg) => pkg.subDirectoryPath);
```

andrewbranch (changes requested):

packages/definitions-parser/src/get-affected-packages.ts
andrewbranch: Ah. I guess you could move the implementation of `get subDirectoryName()` to a utility.

andrewbranch (approved):

packages/definitions-parser/src/get-affected-packages.ts
jakebailey: Yeah though the signals there are different because the real package uses a Boolean for "latest" instead of the star... The whole thing could use some TLC

PR #840, opened 2023-11-21
Add knex dependents to expected failures

gabritto:
DT packages that depend on knex (which ships its own types) have started failing on TS 5.4 due to https://github.com/microsoft/TypeScript/pull/56004.


jakebailey:
`expectedFailures.txt` is read by the package and not via HTTP, so this change will require a changeset (note the failing build).

You should be able to run `pnpm changeset` and mash enter until you can add a changelog entry.

sandersn (approved):

PR #848, opened 2023-11-27
Use package dependencies to manage TypeScript, instead of `$HOME/.dts`

jakebailey:
Rather than `npm install`-ing into `$HOME/.dts`, instead directly depend on all of the typescript versions we support. We already need them all to do anything, so it's no extra work. Now when we drop old versions, they will disappear from user's machines, and we'll no longer clutter `$HOME/.dts`.

`$HOME/.dts` is still used for the suggestions output, though, so we're not all the way there. Would be nice to use a temp dir instead for that.


jakebailey:
This is now ready for review.

andrewbranch (approved):
👏

tsconfig.json
```diff
@@ -10,6 +10,7 @@
     { "path": "packages/header-parser" },
     { "path": "packages/publisher" },
     { "path": "packages/typescript-versions" },
```
sandersn, line 12: this needs to be updated when typescript-versions is updated, right? Can you add that to the instructions in typescript-versions/index.ts?

jakebailey: The list of packages in `package.json` does, yes, though there's a test that enforces that it's up to date so whoever's bumping will definitely know that they forgot. I'll go and add a note, though.

jakebailey: Done.

sandersn (approved):

PR #849, opened 2023-11-28
Useful error on `pnpm test` without arguments

sandersn:
Special-cased for the way `pnpm test` will invoke dtslint on DT.

### DIFF
--- a/packages/dtslint/src/index.ts
+++ b/packages/dtslint/src/index.ts
@@ -26,6 +26,10 @@ async function main(): Promise<void> {
   let tsLocal: string | undefined;
 
   console.log(`dtslint@${require("../package.json").version}`);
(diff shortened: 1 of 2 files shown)


packages/dtslint/src/index.ts
```diff
@@ -26,6 +26,10 @@ async function main(): Promise<void> {
   let tsLocal: string | undefined;
 
   console.log(`dtslint@${require("../package.json").version}`);
+  if (args.length === 1 && args[0] === "types") {
```
jakebailey, line 29: This is maybe not a mega related question, but _why_ do we require the first arg to be `types` at all?

sandersn: The args are just passed on to dtslint joined by `/` I think, so I don't think it's strictly required to be `types`. Not sure.

PR #851, opened 2023-12-01
eslint-plugin: Require eslint ^8.40.0 as peerDependency

hanazuki:
Commit 6da1cb1bfb3e787122536ff14fd09194c8b55cbe started to use `context.sourceCode` property, which is introduced in ESlint v8.40.0.[^eslint-8.40.0]

This patch updates the peerDependency for `eslint` to require that version or later. The change will help users using `@definitelytyped/eslint-plugin` outside of DefinitelyTyped properly manage their dependencies.

[^eslint-8.40.0]: https://eslint.org/blog/2023/05/eslint-v8.40.0-released/

### DIFF
--- a/.changeset/curly-starfishes-check.md
+++ b/.changeset/curly-starfishes-check.md
@@ -1,0 +1,5 @@
+---
+"@definitelytyped/eslint-plugin": patch
+---
+
+Require eslint ^8.40.0 as peerDependency
--- a/packages/eslint-plugin/package.json
+++ b/packages/eslint-plugin/package.json
@@ -29,7 +29,7 @@
   "peerDependencies": {
     "@typescript-eslint/eslint-plugin": "^6.11.0",
     "@typescript-eslint/parser": "^6.11.0",
-    "eslint": "^8.0.0",
+    "eslint": "^8.40.0",
     "eslint-plugin-jsdoc": "^44.0.0",
     "typescript": ">= 3.0.0-dev || >= 3.1.0-dev || >= 3.2.0-dev || >= 3.3.0-dev || >= 3.4.0-dev || >= 3.5.0-dev || >= 3.6.0-dev || >= 3.7.0-dev || >= 3.8.0-dev || >= 3.9.0-dev || >= 4.0.0-dev || >=5.0.0-dev"
   },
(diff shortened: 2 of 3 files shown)


hanazuki:

hanazuki:
I've failed to commit pnpm-lock.yaml and revised patch.

jakebailey (approved):
I'm going to approve this as it's technically correct, but we should really have a discussion about the usability of these rules externally. I am very interested in what you are using this plugin for.

hanazuki:
I'm sorry to be late to reply, and thank you for merging this patch.

My story is that I used to use `dtslint` to lint and type-check the type definitions for a Node.js library I'm maintaining. `dtslint` was a perfect fit for my need because it could type-check type definitions against multiple versions of TypeScript, and had a good ESlint config, although I knew it was created and maintained as an internal tool for DefinitelyTyped project.

As a result of the recent restructuring in DT repository, `dtslint` finally stopped working on external projects. Hence I decided to migrate to another test runner (which might not exist yet) and direct use of ESlint. I found, however, that some of the ESlint rules implemented in this repository (e.g. no-any-union, no-const-enum, redundant-undefined) were still useful for TS type definitions outside DT.

As I understand the packages in this repository are for internal use of the DT project and guarantee no stable versioning/interface, it should be most desirable for me to fork this library and stop depending on DT-tools. But, tracking changes in the TypeScript language features and ESlint internals is equally burdensome for me as tracking DT-tools (or asking to fix DT-tools like this one 🙇). Another option may be to propose typescript-eslint project include these rules, but I guess the rules reside here because they are preferences of the DT project and too opinionated to be there (actually, I don't know how they decide to accept a new rule in their default ruleset).

I also think the ESLint config from DT/eslint-plugin can be valuable for library authors maintaining their type definitions in their repository because their users and contributors are likely familiar with the DT coding convention and it can provide a consistent experience.

jakebailey:
There's definitely still a wider question of if we expect people to be able to run this repo's code outside of DT, though I will note that `no-const-enum` could be replaced with `no-restricted-syntax`. `no-any-union` and `redundant-undefined` are certainly ones that would be useful in ts-eslint, IMO.

jakebailey:
Ah, ts-eslint already has https://typescript-eslint.io/rules/no-redundant-type-constituents/ which covers `no-any-union` though does so with type info (and checks for more badness). I'm going to check to see if we could just use this.

hanazuki:
@jakebailey Thank you for your advice on ESlint rules.

PR #865, opened 2023-12-12
Add new no-type-only-packages lint rule

jakebailey:
This PR adds a new lint rule which bans packages which do not contribute any value types, with some exceptions:

- Interface merging with a type declared outside of the package. This covers extending global interfaces like `Window`, extending someone else's package, etc.
- Fully empty packages (so starting with nothing doesn't complain)

I will run this against DT to produce a full list of packages which currently fail (which has caught bugs in this rule I threw together).


jakebailey:
Currently failing packages, excluding those which I believe should fail:

- [ ] Do `export * from ...`
	- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/draco3dgltf/index.d.ts
	- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/ramda/index.d.ts
	- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/redis-mock/index.d.ts
	- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/uglify-es/index.d.ts
	- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/wordpress__block-serialization-default-parser/index.d.ts
	- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/yog-ral/index.d.ts
- [ ] Have files outside the main program that have values
	- https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/freedom
- [ ] Extending `Window` wasn't ignore d work
	- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/plaid-link/index.d.ts
- [ ] No idea
	- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/keystonejs__session/index.d.ts

packages/eslint-plugin/test/__file_snapshots__/types/no-useless-files/index.d.ts.lint
```diff
@@ -1,5 +1,10 @@
-No errors
+types/no-useless-files/index.d.ts
+  1:1  error  Packages should contain value components, not just types  @definitelytyped/no-type-only-packages
+
+✖ 1 problem (1 error, 0 warnings)
 
 ==== types/no-useless-files/index.d.ts ====
 
     export default "I am useful";
```
sandersn, line 8: this looks like a value to me. Am I wrong?

sandersn (changes requested):
Couple of questions about the tests.

packages/eslint-plugin/test/__file_snapshots__/types/no-useless-files/index.d.ts.lint
jakebailey: Yep, this is a bug. I didn't notice.

packages/eslint-plugin/test/__file_snapshots__/types/no-bad-reference/v11/index.d.ts.lint
```diff
@@ -1,4 +1,5 @@
 types/no-bad-reference/v11/index.d.ts
+   1:1   error  Packages should contain value components, not just types                                                                     @definitelytyped/no-type-only-packages
```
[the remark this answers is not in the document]
jakebailey: If you're only comments, then the rule doesn't fire. But if you ever add a statement, including an import or export, then we check.

It's possible that an import brings in a global, but if we consider that to be true then we basically won't error anywhere. I haven't come up with a better heuristic, but maybe there's something I've missed.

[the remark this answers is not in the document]
sandersn: I think this is good enough, I just wanted to double check.

packages/eslint-plugin/test/__file_snapshots__/types/no-useless-files/index.d.ts.lint
jakebailey: Fixed.

sandersn (approved):

jakebailey:
Should I just disable this lint for all packages which fail it currently, under the assumption that it's too late to "reject" them (per se)?

sandersn:
Yes. It'll be good to have that list easily obtainable by `grep no-type-only types/*/.eslintrc.json` though.

jakebailey:
I've verified that this PR does not seem to negatively impact performance. _Maybe_ 3 minutes extra on a full run (`1h17m` -> `1h20m`), but that was observed on different days.

The performance problem I had was specifically to having `pnpm link`'d which caused multiple `typescript-eslint/estree` packages to be in the workspace. I've added code that can handle this case (for our own dev work), but it would not happen on DT normally. In the future, the "flat config" method will allow us to avoid all of that stuff.

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/68180 is ready and disables the lint.

PR #867, opened 2023-12-13
Remove performance logging

jakebailey:
I removed the code that generated these files in #843; maybe I should restore it, but it's not super clear to me we can do so now that we're in eslint and running per-file and not a full compilation.


sandersn (approved):
It might be possible to rescue this with some cleverness, but at this point I'd rather write a one-off script to run locally instead.

PR #866, opened 2023-12-12
feat: remove TSLint

JoshuaKGoldberg:
Fixes #648! 🥳🔪

Now that all the old TSLint rules have been switch to ESLint equivalents, there's no more need for TSLint in `dtslint`. This PR removes the `tslint` package reference and tooling around it.

PR #871, opened 2023-12-13
fix(header-parser): remove tsutils exemption from getTypesVersions

JoshuaKGoldberg:
Forking from https://github.com/microsoft/DefinitelyTyped-tools/pull/866#discussion_r1424620621: we only have educated guesses as to why this exemption exists.


jakebailey (approved):
This needs to merge main, but otherwise I think this is fine.

This code was added in https://github.com/microsoft/DefinitelyTyped-tools/commit/feb21c960a9afbd034655add42eafcbd93051bcd, but there's no explanation as to why.

sandersn:
Well, good news, there is no folder named tsutils on DT anyway. Looks like it was a folder in graphql, which was deprecated 4 years ago.

sandersn (approved):

sandersn:
Added changeset to make the error go away, even though this doesn't really need to ship right away. It's a little weird to have it error, but I guess it's a good reminder. @jakebailey is the intent for every PR to have a changeset entry, or only things that need to ship right away?

jakebailey:
Everything needs a changeset if it touches code, because otherwise the changeset publishing tool won't know to publish something when another package may depend on the behavior. The only "okay" time is if the change is test only or you definitely don't care (like the comment only changes in the tslint removal PR).

This PR definitely should have a changeset.

PR #875, opened 2023-12-14
Add dts-gen to repo

jakebailey:
This pulls https://github.com/microsoft/dts-gen into the repo, with these changes:

- Remove the prototype browser stuff (which wasn't touched in years)
- Don't baseline builtin node packages
- Baselines are done via jest versus bespoke implementation
- Includes https://github.com/microsoft/dts-gen/pull/192


packages/dts-gen/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "@types/parse-git-config": "^1.1.0",
+    "@types/yargs": "latest",
+    "ecurve": "1.0.5",
+    "jquery": "3.7.1",
+    "lodash": "4.17.21"
+  },
+  "keywords": [
+    "typescript"
+  ],
+  "preferGlobal": true
```
jakebailey, line 43: I have no idea what this is but it was there in the original package.

packages/dts-gen/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "bugs": {
+    "url": "https://github.com/microsoft/DefinitelyTyped-tools/issues"
+  },
+  "engines": {
+    "node": ">=16.17.0"
+  },
+  "dependencies": {
+    "dts-dom": "^3.7.0",
+    "parse-git-config": "^1.1.1",
+    "typescript": "^3.5.1",
```
jakebailey, line 29: This is ancient. I haven't modified this but I am definitely curious if we can safely bump this.

sandersn (changes requested):
Looks reasonable except that the version needs to be 0.7.4 now, and I would like to know what versioning scheme we'll be using in the future.

packages/dts-gen/package.json
jakebailey: It's a real dep; it uses TS to parse and walk ASTs and so on. But I suspect that this being ancient means that it's not been working super well.

packages/dts-gen/package.json
jakebailey: Apparently it's https://docs.npmjs.com/cli/v6/configuring-npm/package-json#preferglobal which is deprecated. I'll drop it here.

packages/dts-gen/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "dts-gen",
+  "version": "0.7.3",
```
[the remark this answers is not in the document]
jakebailey: It should, but I'll bump it to 0.7.4.

Right now on merge and release, it'll go to 0.8.0, though. We could do a major bump if wanted; since it's zero versioned, there's no difference.

packages/dts-gen/package.json
jakebailey: Bumped to 5.3.3. FWIW the repo in general is still on 5.2.2, which we should update.

packages/dts-gen/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "typescript": "^5.3.3",
+    "yargs": "^4.8.1"
+  },
+  "devDependencies": {
+    "@types/node": "^16.18.61",
+    "@types/parse-git-config": "^1.1.0",
+    "@types/yargs": "latest",
+    "ecurve": "1.0.5",
+    "jquery": "3.7.1",
+    "lodash": "4.17.21"
```
jakebailey, line 38: These deps are here solely for tests. Hopefully none of these have CVEs or something.

sandersn (approved):

PR #879, opened 2023-12-14
Clean up eslint options building

jakebailey:
This new method should be less error prone.

### DIFF
--- a/packages/dtslint/src/lint.ts
+++ b/packages/dtslint/src/lint.ts
@@ -48,8 +48,8 @@ export async function lint(
   const options = getEslintOptions(expectOnly, minVersion, maxVersion, tsLocal);
   const eslint = new ESLint(options);
   const formatter = await eslint.loadFormatter("stylish");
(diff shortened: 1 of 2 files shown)


packages/dtslint/src/lint.ts
```diff
@@ -44,58 +44,68 @@ export async function lint(
       files.push(fileName);
     }
   }
-  let output = "";
 
+  const options = getEslintOptions(expectOnly, minVersion, maxVersion, tsLocal);
+  const eslint = new ESLint(options);
+  const formatter = await eslint.loadFormatter("stylish");
+  const eresults = await eslint.lintFiles(esfiles);
```
sandersn, line 51: looks like this was moved before the esfiles->files rename

sandersn (approved):

packages/dtslint/src/lint.ts
jakebailey: Yeah, bad merge.

PR #880, opened 2023-12-14
Update knip

jakebailey:


jakebailey:
Theoretically https://knip.dev/reference/configuration/#ignoredependencies should work, but does not. https://github.com/webpro/knip/issues/392

.npmrc
```diff
@@ -1,4 +1,3 @@
 prefer-workspace-packages=true
-node-version=16.17.0
```
jakebailey, line 2: I put this in here for safety, but since knip needs 18+, this doesn't work.

sandersn (approved):

PR #884, opened 2023-12-15
Make dts-gen private instead

jakebailey:
The past PR didn't work, no clue why ignore didn't actually ignore. private=true should work...


jakebailey:
The changeset check is red, but I can't introduce a new changeset as that will stop the publish of other packages too. This is more of a lint check and PRs don't need changesets always so I'm going to just merge it :(

PR #891, opened 2023-12-24
feat(eslint-plugin): Enabled parserOptions.allowAutomaticSingleRunInference

### LINKED ISSUE #890: Enable parserOptions.allowAutomaticSingleRunInference
Coming over from https://twitter.com/andhaveaniceday/status/1738986013320794313: typescript-eslint has a relatively undocumented `allowAutomaticSingleRunInference` feature that in many cases speeds up single-run linting. That includes CI.

Because https://github.com/eslint/rfcs/pull/102 wasn't accepted (😞) and one last bug with it has yet to be explained (https://github.com/typescript-eslint/typescript-eslint/issues/3851), we've been cautious to enable it by default in typescript-eslint. But as @jakebailey noted, [DT doesn't] have project references and it's all single run outside of the editor.

JoshuaKGoldberg:
Fixes #890.

### DIFF
--- a/.changeset/quick-beds-cheer.md
+++ b/.changeset/quick-beds-cheer.md
@@ -1,0 +1,5 @@
+---
+"@definitelytyped/eslint-plugin": patch
+---
+
+Enabled parserOptions.allowAutomaticSingleRunInference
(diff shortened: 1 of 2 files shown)


jakebailey:
I think we're leaving #890 open since we want to understand why this option isn't helping much, but I believe this PR can be closed given we already enable it via the environment variable.

PR #895, opened 2024-01-04
Clean up expect type code for diagnostics

jakebailey:
This makes the followup changes we discussed for #892.

Namely, determine if diagnostics should be reported first, and then pass that down so we can bail early. This should speed things up.

We don't need the precheck for being a dts file at all; `walk` does this immediately after reporting diagnostics.

I've also added a test; these fixture tests are pretty nice for being able to test different options with multiple real eslint configs.


packages/eslint-plugin/test/__file_snapshots__/types/expect-diagnostics/index.d.ts.lint
```diff
@@ -0,0 +1,8 @@
+No errors
+
+==== types/expect-diagnostics/index.d.ts ====
+
+    export const foo = 1234;
+    
+    // This should cause a diagnostic.
```
sandersn, line 7: no diagnostic here

packages/eslint-plugin/test/__file_snapshots__/types/expect-diagnostics/expect-diagnostics-tests.ts.lint
```diff
@@ -0,0 +1,9 @@
+No errors
+
+==== types/expect-diagnostics/expect-diagnostics-tests.ts ====
+
+    // eslint-disable-next-line @definitelytyped/no-relative-import-in-test
+    import * as expect from ".";
+    
+    // This should cause a diagnostic.
```
sandersn, line 8: no diagnostic here

packages/eslint-plugin/test/fixtures/types/expect-diagnostics-versioned/.eslintrc.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  overrides: [
+    {
+      files: ["*.ts", "*.cts", "*.mts", "*.tsx"],
+      rules: {
+        "@definitelytyped/expect": ["error", { versionsToTest: [{ versionName: "x.y", path: "typescript" }] }],
```
sandersn, line 6: how does this test work? Does it treat "x.y" as a real Typescript version somehow?

sandersn (changes requested):
Code *looks* right, but the unversioned tests don't.

packages/eslint-plugin/test/fixtures/types/expect-diagnostics-versioned/.eslintrc.js
jakebailey: This is just a pretty name, the ones we generate when creating this list in `dtslint`. It's used instead of the version provided by `typescript`. Here I'm just giving it a random name since I don't want to hardcode/reuse the version from the current repo as it changes (and would break baselines).

packages/eslint-plugin/test/__file_snapshots__/types/expect-diagnostics/index.d.ts.lint
jakebailey: Renamed and changed the comment, thanks for noticing.

sandersn (approved):

PR #899, opened 2024-01-10
Skip only npm naming in dependents

sandersn:
Previously, everything was skipped except for expect. This will be
slower, but will catch more errors. Now that we're running only eslint,
it should be fast enough, but I'll test with a node PR after this ships.

### DIFF
--- a/packages/dtslint-runner/src/main.ts
+++ b/packages/dtslint-runner/src/main.ts
@@ -62,7 +62,7 @@ export async function runDTSLint({
   await runWithListeningChildProcesses({
     inputs: testedPackages.map((path) => ({
       path,
-      onlyTestTsNext: onlyTestTsNext || !packageNames.has(path),
+      onlyTestTsNext,
       expectOnly,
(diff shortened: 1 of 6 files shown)


packages/dtslint-runner/src/main.ts
```diff
@@ -62,7 +63,8 @@ export async function runDTSLint({
     inputs: testedPackages.map((path) => ({
       path,
       onlyTestTsNext: onlyTestTsNext || !packageNames.has(path),
```
jakebailey, line 65:
```suggestion
      onlyTestTsNext,
```

I feel like this should change too; it seems strange that we'd _enable_ testing of non-next versions when testing only next on the changed packages?

jakebailey (approved):
LGTM with one comment.

packages/dtslint-runner/src/main.ts
sandersn: Good catch -- that is pretty nonsensical.

PR #903, opened 2024-01-12
Loosen checkParseResults

sandersn:
1. Don't require types dependencies to be on DT.
2. Allow types dependencies to be pinned.
3. Don't bother asserting that dependencies from package.json are in fact present in package.json.

The remaining check requires dependents to have a higher-or-equal minTypescriptVersion compared to their dependency. This should really be in dtslint, but I don't want to make the fix (too) complex. And I haven't gone to check whether dtslint has access to `allPackages`.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -223,16 +223,10 @@ importers:
       '@octokit/rest':
         specifier: ^19.0.13
         version: 19.0.13
-      semver:
-        specifier: ^7.5.4
-        version: 7.5.4
       yargs:
         specifier: ^17.7.2
         version: 17.7.2
     devDependencies:
-      '@types/semver':
-        specifier: ^7.5.5
-        version: 7.5.5
       glob:
         specifier: ^10.3.10
         version: 10.3.10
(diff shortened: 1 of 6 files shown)


pnpm-lock.yaml
```diff
@@ -4969,7 +4963,7 @@ packages:
       pretty-ms: 8.0.0
       strip-json-comments: 5.0.1
       summary: 2.1.0
-      typescript: 5.3.3
+      typescript: 5.2.2
```
jakebailey, line 4966: Hm, what happened here? Why is this not 5.3.3?

jakebailey: Super odd:

```
/knip@3.8.1(@types/node@16.18.61)(typescript@5.2.2):
```

Probably we should upgrade TS in everything anyway.

jakebailey (approved):

pnpm-lock.yaml
sandersn: I ran `pnpm i` and the top-level package.json has 5.2.2. I can change it to 5.3.3 I guess.

jakebailey: No worries, I don't think it's a problem. My other PR updates knip anyway.

pnpm-lock.yaml
```diff
@@ -2083,6 +2083,36 @@ packages:
       typescript: 5.2.2
     transitivePeerDependencies:
       - supports-color
+    dev: false
+
+  /@typescript-eslint/eslint-plugin@6.11.0(@typescript-eslint/parser@6.11.0)(eslint@8.53.0)(typescript@5.3.3):
```
jakebailey, line 2088: Oof, this isn't going well. I'd just leave your PR as it was and we can do a big dep update elsewhere.

PR #907, opened 2024-01-16
Handle moves in gitChanges+other cleanup

sandersn:
1. gitDiff parses git moves from R*nnn* *from* *to* to `{ status: "R", file: to, source: from }`.
1. gitChanges converts `R` to an add and a remove.
3. Documentation of gitChanges, especially in relation getAffectedPackages.
4. Test of gitChanges!
5. No more circular dependency get-affected-packages.ts <-> git.ts. Now git.ts only depends on get-affected-packages.ts.
6. Dedupe call to (and error logging from!) gitChanges.
7. getAffectedPackages no longer returns any errors, only throws if `pnpm` execs fail.

The test case I added is pretty big, but it is real, and complicated. I can try to minimise it if anybody feels strongly that it should be.

### DIFF
--- a/packages/definitions-parser/src/git.ts
+++ b/packages/definitions-parser/src/git.ts
@@ -6,10 +6,16 @@ import * as semver from "semver";
 import { inspect } from "util";
 import { PreparePackagesResult, getAffectedPackages } from "./get-affected-packages";
 
-export interface GitDiff {
-  status: "A" | "D" | "M";
-  file: string;
-}
+export type GitDiff =
+  | {
+      status: "A" | "D" | "M";
+      file: string;
+    }
+  | {
+      status: "R";
+      file: string;
+      source: string;
+    };
 
 /*
 We have to be careful about how we get the diff because Actions uses a shallow clone.
@@ -38,7 +44,10 @@ export async function gitDiff(log: Logger, definitelyTypedPath: string): Promise
     diff = (await run("git", ["diff", `${sourceBranch}~1`, "--name-status"])).trim();
   }
   return diff.split("\n").map((line) => {
-    const [status, file] = line.split(/\s+/, 2);
+    const [status, file, destination] = line.split(/\s+/, 3);
+    if (status[0] === "R") {
+      return { status: "R", file: destination.trim(), source: file.trim() };
+    }
     return { status: status.trim(), file: file.trim() } as GitDiff;
   });
 
@@ -50,6 +59,10 @@ export async function gitDiff(log: Logger, definitelyTypedPath: string): Promise
   }
 }
 
+/**
+ * @returns packages with added or removed files, but not packages with only changed files; 
+ * {@link getAffectedPackages | those are found by calling pnpm }.
+ */
 export function gitChanges(
   diffs: GitDiff[],
 ): { errors: string[] } | { deletions: PackageId[]; additions: PackageId[] } {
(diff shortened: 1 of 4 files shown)


packages/definitions-parser/src/git.ts
```diff
@@ -148,11 +174,8 @@ it is supposed to replace, ${typings.version} of ${unneeded.name}.`);
  */
 export async function getNotNeededPackages(
   allPackages: AllPackages,
-  diffs: GitDiff[],
+  deletions: PackageId[],
```
jakebailey, line 177: At some level, it feels weird that a getter is also doing these kinds of checks, but not really for fixing in this PR.

packages/definitions-parser/src/git.ts
```diff
@@ -61,12 +74,24 @@ export function gitChanges(
...
+          const srcKey = `${srcDep.typesDirectoryName}/v${
+            srcDep.version === "*" ? "*" : formatTypingVersion(srcDep.version)
+          }`;
+          addedPackages.set(srcKey, [srcDep, "D"]);
+        }
+      } else {
+        addedPackages.set(key, [dep, diff.status]);
+      }
     } else {
+      const status = diff.status === "A" || diff.status === "R" ? "add" : "delete";
```
jakebailey, line 90:
```suggestion
      const status = diff.status === "A" || diff.status === "D" ? "add" : "delete";
```

I think this is a typo? This else branch can't have `R`, right?

sandersn: The else branch just above this one can't; this is the else branch when the file being moved outside the monorepo structure. That is, for files that's don't match the pattern `types/X/file.ts`

jakebailey: Oh oops, I misread this as `A -> add, R -> delete`, not `A or R -> add, else delete`.

packages/definitions-parser/src/git.ts
sandersn: Yeah, the function is primarily "check the packages of these files" but it also returns a list of those packages that don't have errors.
Not sure changing the name would make it better, though, because then it would be a check- function that returns something.

packages/definitions-parser/test/git.test.ts
```diff
@@ -21,43 +21,104 @@ const deleteJestDiffs: GitDiff[] = [
...
+  async gitMovesConvertedToAddsAndDeletes() {
+    expect(gitChanges(moveRdfJSDiffs)).toEqual({
+      // serializer-jsonld-ext has changes, why isn't in the list? I guess we only track dirs with additions and deletions?
+      additions: [
+        { typesDirectoryName: "rdf-ext", version: "*" },
+        { typesDirectoryName: "rdfjs__formats", version: "*" },
+        { typesDirectoryName: "rdfjs__serializer-turtle", version: "*" },
+      ],
+      deletions: [
+        { typesDirectoryName: "rdfjs__environment", version: "*" }, // this still has stuff in it, why is it put in deletions?
```
jakebailey, line 82: RE: this comment and the other comment above, the diffs only track what changed, yeah. I think that if anything, these could be named "potential additions" and "potential deletions" and we already run tests assuming those are the case?

Maybe this means that we are missing some check for "deleted some of the files but not all" where it needs to at least see if the package seems like it exists... but I think that's something you already fixed?

jakebailey (approved):

packages/definitions-parser/test/git.test.ts
sandersn: Oh, I need to remove this question; it is indeed "packages *with* deleted *files*", not "deleted *packages*".

jakebailey (approved):

PR #908, opened 2024-01-16
Make node16 check in expect case insensitive

jakebailey:
This was caught on https://github.com/DefinitelyTyped/DefinitelyTyped/pull/68126. Copy the check from https://github.com/microsoft/DefinitelyTyped-tools/blob/4ec06280828fbd22c48008dfbf423585ab8a2d87/packages/dtslint/src/checks.ts#L83.


andrewbranch (approved):
Oops, thanks for figuring this out!

PR #910, opened 2024-01-16
Require Node 18+

jakebailey:
Perhaps we should minor/major bump for this; since we only target DT itself I'm not sure if we actually care.


andrewbranch:
FWIW I’m not attached to any particular versioning scheme here. We can do our first minor bump if you want 😄

andrewbranch (approved):

jakebailey:
> FWIW I’m not attached to any particular versioning scheme here. We can do our first minor bump if you want 😄

I'm afraid that already happened accidentally in #866 ☹️

PR #913, opened 2024-01-17
Write suggestions to the cache dir instead of home

### LINKED ISSUE #387: Use the correct XDG directory instead of putting a directory in the user home
Programs (especially new ones) that wants to fit in Linux should use the XDG dir standard instead of just dropping a new directory in the user home. The standard can be seen here: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

It would be nice to move the `.dts` directory to its appropriate place

remcohaszing:
This prevents DefinitelyTyped tools from polluting the user’s home directory.

Also the `suggestionsDir` is defined and exported once, so it can be reused.

Closes #387

### DIFF
--- a/.changeset/funny-ravens-carry.md
+++ b/.changeset/funny-ravens-carry.md
@@ -1,0 +1,5 @@
+---
+"@definitelytyped/utils": patch
+---
+
+Change suggestionsDir to use the user’s cache dir instead of .dts the user’s home dir
--- a/.changeset/tiny-boats-prove.md
+++ b/.changeset/tiny-boats-prove.md
@@ -1,0 +1,5 @@
+---
+"@definitelytyped/utils": patch
+---
+
+Expose suggestionsDir
--- a/packages/dtslint-runner/src/main.ts
+++ b/packages/dtslint-runner/src/main.ts
@@ -1,16 +1,14 @@
-import os from "os";
 import {
   execAndThrowErrors,
   joinPaths,
   runWithListeningChildProcesses,
+  suggestionsDir,
   CrashRecoveryState,
 } from "@definitelytyped/utils";
 import fs from "fs";
 import { RunDTSLintOptions } from "./types";
 import { prepareAllPackages } from "./prepareAllPackages";
 import { prepareAffectedPackages } from "./prepareAffectedPackages";
-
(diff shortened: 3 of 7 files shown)


jakebailey:
This path is known by CI so dangerbot can report suggestions, so if this changes, that has to change too.

remcohaszing:
I think you’re talking about [this line](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/942c50bb4d57679ca1dcb210f4984b41e7bbfdbb/dangerfile.ts#L6), right? It could even `import { suggestionsDir } from '@definitelytyped/utils' after this change.

The change could also be done in three steps to be safe:

1. Define `suggestionsDir` in `@definitelytyped/utils` and reuse it.
2. Use import `suggestionsDir` from `@definitelytyped/utils` in `dangerfile.ts`
3. Change `suggestionsDir` to use `cachedir`.

jakebailey:
Ah, yeah, we could just import it, though I'm unsure if that will work as the dangerfile is not directly executed, but more or less eval'd (so, not sure if an ESM import will actally work!)

remcohaszing:
So that would mean it can’t use any dependencies, not even `cachedir`. Is `dangerfile.ts` even executed outside CI? Otherwise the path can be more-or-less hardcoded to this line: https://github.com/LinusU/node-cachedir/blob/master/index.js#L5.

jakebailey:
> So that would mean it can’t use any dependencies, not even `cachedir`. Is `dangerfile.ts` even executed outside CI? Otherwise the path can be more-or-less hardcoded to this line: [LinusU/node-cachedir@`master`/index.js#L5](https://github.com/LinusU/node-cachedir/blob/master/index.js?rgh-link-date=2024-01-19T14%3A40%3A02Z#L5).

Yep, you're right; I'm just plain wrong. I used `strip-ansi` in the dangerfile before and it was fine, so I'm clearly not remembering clearly.

jakebailey:
With that in mind, we can just make a PR in the DT repo that includes the code you describe (import the util) and then we can release this one and verify that the other one works.

(We could do two releases to make it most definitely break-free, but, everything depends on utils and will get bumped so that's maybe slightly wasteful. But, not bad either.)

jakebailey (approved):

PR #920, opened 2024-01-29
Require one .d.ts, one .ts in tsconfig

sandersn:
Previously this was checked implicitly by the unused-files check. Now
it's handled only by .npmignore, which works for publishing but not for
CI -- in particular, it's easy to forget to put test files into
tsconfig, thereby skipping testing.

This fix is NOT foolproof, but it will prevent common mistakes while
allowing uncommon layouts. `.npmignore` will continue to make sure that stray files
don't get published.

### DIFF
--- a/packages/dtslint/src/checks.ts
+++ b/packages/dtslint/src/checks.ts
@@ -3,7 +3,7 @@ import { AllTypeScriptVersion } from "@definitelytyped/typescript-versions";
 import fs from "fs";
 import { join as joinPaths } from "path";
 import { CompilerOptions } from "typescript";
-import { deepEquals } from "@definitelytyped/utils";
+import { deepEquals, isDeclarationPath } from "@definitelytyped/utils";
 
 import { readJson, packageNameFromPath } from "./util";
 export function checkPackageJson(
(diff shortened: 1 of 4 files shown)


packages/dtslint/src/checks.ts
```diff
@@ -24,14 +24,34 @@ export type CompilerOptionsRaw = {
...
+  const options = config.compilerOptions;
+  if ("include" in config) {
+    errors.push('Use "files" instead of "include".');
+  }
+  if ("exclude" in config) {
+    errors.push('Use "files" instead of "exclude".');
+  }
+  if (!config.files) {
+    errors.push('Must specify "files".');
+  } else if (!config.files.some((f) => f.endsWith(".d.ts"))) {
```
jakebailey, line 50: I was going to suggest that you use `isDeclarationPath` here, which is _technically_ the right answer, but I actually think that we should be checking to verify that `index.d.ts` in particular is present (with a todo for `isDeclaraionPath`).

The reason is that there are two eslint rules which only run on the "main" file, which is hardcoded `index.d.ts`, and they both rely on being able to ask eslint for the Program based on that file (so, `index.d.ts` must be in the Program).

packages/dtslint/src/checks.ts
```diff
@@ -24,14 +24,34 @@ export type CompilerOptionsRaw = {
...
+    errors.push('"files" list must include at least one ".d.ts" file.');
+  } else if (!config.files.some((f) => f.endsWith(".ts") && !f.endsWith(".d.ts"))) {
```
jakebailey, line 52: Theoretically this should check for `.mts` and `.cts` too. Probably like:

```ts
!config.files.some((f) => /\.[cm]?ts$/.test(f) && !isDeclarationPath(f))
```

packages/dtslint/test/index.test.ts
```diff
@@ -14,98 +14,136 @@ describe("dtslint", () => {
     noEmit: true,
     forceConsistentCasingInFileNames: true,
   };
+  function based(extra: object) {
```
jakebailey, line 17: what a function name

jakebailey (approved):
LGTM; curious what tsconfigs currently fail.

sandersn:
It should be a small number because we had unused file detection until .npmignore I believe, so about two months worth of changes.

packages/dtslint/test/index.test.ts
sandersn: I just write whatever bubbles out of my subconscious.

PR #943, opened 2024-02-08
Run arethetypeswrong from in-memory tarball data

andrewbranch:
This should be a performance win, but is also preparation for fixing https://github.com/DefinitelyTyped/DefinitelyTyped/issues/67361 and bringing dts-critic suggestions back.


jakebailey (approved):

jakebailey:
Needs a changeset, but, after that I'm excited for in-memory. Would love to eliminate all of our other calls out to npm for info too.

PR #944, opened 2024-02-08
Detect package names added/removed from attw.json as changed

andrewbranch:
For https://github.com/DefinitelyTyped/DefinitelyTyped/pull/68558 (and others in the future)

### DIFF
--- a/packages/dtslint/src/index.ts
+++ b/packages/dtslint/src/index.ts
@@ -21,6 +21,7 @@ async function main(): Promise<void> {
   let onlyTestTsNext = false;
   let expectOnly = false;
   let skipNpmChecks = false;
+  let onlyNpmChecks = false;
   let shouldListen = false;
   let lookingForTsLocal = false;
   let tsLocal: string | undefined;
@@ -55,6 +56,9 @@ async function main(): Promise<void> {
       case "--skipNpmChecks":
         skipNpmChecks = true;
         break;
+      case "--onlyNpmChecks":
+        onlyNpmChecks = true;
+        break;
       case "--onlyTestTsNext":
         onlyTestTsNext = true;
         break;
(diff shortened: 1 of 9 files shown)


jakebailey:
The way that we install things means that this will cause new failures; we use a filtered install so pnpm is not going to know that it needs to install those packages too and so they'll definitely fail tests.

A potential way we can try and do this is to update https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/scripts/pnpm-install.sh to detect that case and tack on a few more filters.

That or we move `pnpm install` back to `dtslint-runner`, but there are other reasons I want to try and avoid that.

packages/dtslint/src/index.ts
```diff
@@ -160,38 +174,40 @@ async function runTests(
...
-      const hi = his[i];
-      assert(
-        parseFloat(hi) >= parseFloat(low),
-        `'"minimumTypeScriptVersion": "${minVersion}"' in package.json skips ts${hi} folder.`,
-      );
-      const isLatest = hi === TypeScriptVersion.latest;
-      const versionPath = isLatest ? dirPath : joinPaths(dirPath, `ts${hi}`);
-      if (lows.length > 1) {
-        console.log("testing from", low, "to", hi, "in", versionPath);
+  if (npmChecks !== "only") {
```
jakebailey, line 184: Maybe this is a nit, but if this were written as an early return, I think it may be clearer / less diff-y?

```ts
if (npmChecks === "only") {
   return;
}
```

andrewbranch: There’s stuff after this block, though

jakebailey: Ah, crap. The GitHub diff ended perfectly right after the block and I really thought the function had ended 😄 

image

jakebailey (approved):
Seems reasonable now that it will just run ATTW on those packages removed but unchanged. I have not tested it on a specific PR, though.

andrewbranch:
I tested it locally on https://github.com/DefinitelyTyped/DefinitelyTyped/pull/68558 and it worked.

andrewbranch:
Working IRL: https://github.com/DefinitelyTyped/DefinitelyTyped/actions/runs/7837507900/job/21387317058?pr=68558

PR #956, opened 2024-02-17
fix: remove `aws-lambda` from expected npm version failures

G-Rath:
See https://github.com/DefinitelyTyped/DefinitelyTyped/pull/68667 - I think I'm doing this right...

### DIFF
--- a/.changeset/eleven-emus-warn.md
+++ b/.changeset/eleven-emus-warn.md
@@ -1,0 +1,5 @@
+---
+"@definitelytyped/dtslint": patch
+---
+
+remove `aws-lambda` from expected npm version failures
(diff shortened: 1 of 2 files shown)


andrewbranch:
Yep, thank you!

.changeset/eleven-emus-warn.md
```diff
@@ -0,0 +1,5 @@
+---
+"@definitelytyped/dtslint": patch
+---
+
+remove `aws-lambda` from expected npm version failures
```
jakebailey, line 5: Is this file pulled from the repo or from the package? If from the repo, we should update the workflow to not require this.

andrewbranch: I think it's from the package right now. Would be nice to do it from the repo.

jakebailey: Going through my email backlog from being out; I'll send a PR for this quick!

jakebailey: Sent https://github.com/microsoft/DefinitelyTyped-tools/pull/959

PR #959, opened 2024-02-21
 Pull expectedNpmVersionFailures from GitHub, like allowedPackageJsonDependencies

jakebailey:
~Perhaps this is a bit too copy-pasty; the two functions are very similar and I could extract them into a common function, but it seemed somewhat awkward thanks to the differing paths.~


jakebailey:
Actually, not that bad, I think; will refactor.

andrewbranch (approved):
Thanks!

PR #968, opened 2024-02-29
Error on unused vars in declaration files

jakebailey:
This is a replacement for https://github.com/DefinitelyTyped/DefinitelyTyped/pull/44272, aiming to complain in declaration files only.

Running this on all of DT now to see what happens, but I suspect that it's a lot of double-firing for "I used a type parameter one time".


jakebailey:
Stopped my run once I noticed https://github.com/typescript-eslint/typescript-eslint/issues/2867; that'll need to be fixed first.

PR #973, opened 2024-03-05
Support multiple tsconfig files in expect

jakebailey:
To make this work, a package can configure package.json with

```json
{
	"tsconfigs": ["tsconfig.non-dom.json", "tsconfig.dom.json"]
}
```

These must be named like `tsconfig.*.json`. They must not start with `./` or anything and must be present in `tsX.Y` dirs.

This still works less than optimally because ts-eslint needs to be able to match a tsconfig for _every_ file, which means the main `tsconfig` still has to include every file. This means that if you want a tsconfig that _excludes_ a file, you need three tsconfigs in order to still have one with everything.

e.g., node has DOM and non-dom; if you want to test both without using expect's `||`, then you need to have two differing test files, but then that means three tsconfigs: one with both, then two with one of each.

TODO:

- [ ] dt-mergebot needs to know about these files somehow
	- Just check for `tsconfig.*.json`.
- [x] Is this really the right configuration method? Not package.json?

### DIFF
--- a/packages/dtslint/src/index.ts
+++ b/packages/dtslint/src/index.ts
@@ -195,6 +195,7 @@ async function runTests(
       const tsVersion = tsLocal ? "local" : TypeScriptVersion.latest;
       const testTypesResult = await testTypesVersion(
         dirPath,
+        packageJson.tsconfigs,
         tsVersion,
         tsVersion,
         expectOnly,
@@ -222,7 +223,7 @@ async function runTests(
         if (lows.length > 1) {
           console.log("testing from", low, "to", hi, "in", versionPath);
         }
-        const testTypesResult = await testTypesVersion(versionPath, low, hi, expectOnly, undefined, isLatest);
+        const testTypesResult = await testTypesVersion(versionPath, packageJson.tsconfigs, low, hi, expectOnly, undefined, isLatest);
         errors.push(...testTypesResult.errors);
       }
     }
@@ -264,6 +265,7 @@ function next(v: TypeScriptVersion): TypeScriptVersion {
 
 async function testTypesVersion(
   dirPath: string,
+  tsconfigs: readonly string[],
   lowVersion: TsVersion,
   hiVersion: TsVersion,
   expectOnly: boolean,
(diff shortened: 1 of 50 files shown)


jakebailey:
Updated to not use settings, and instead specify tsconfigs in package.json. For sake of dt-mergebot, I'm also going to update things to say that the tsconfigs need to have a specific prefix if not named tsconfig.json.

packages/dtslint/src/index.ts
```diff
@@ -273,10 +275,15 @@ async function testTypesVersion(
...
   errors.push(...checkExpectedFilesResult.errors);
-  const tsconfigErrors = checkTsconfig(dirPath, getCompilerOptions(dirPath));
-  if (tsconfigErrors.length > 0) {
-    errors.push("\n\t* " + tsconfigErrors.join("\n\t* "));
+
+  for (const tsconfig of tsconfigs) {
+    const tsconfigPath = joinPaths(dirPath, tsconfig);
+    const tsconfigErrors = checkTsconfig(dirPath, getCompilerOptions(tsconfigPath));
+    if (tsconfigErrors.length > 0) {
+      errors.push("\n\t* " + tsconfigErrors.join("\n\t* "));
```
andrewbranch, line 283: Is it clear from the error text which tsconfig this applies to?

andrewbranch (approved):

packages/dtslint/src/index.ts
jakebailey: Nope, definitely not, will fix.

jakebailey: Modified this to include a filename before each of these groups.

jakebailey:
I'm going to hold off on merging this because I want to finish a PR that would work for Node.

jakebailey:
This is all working; I have not done the dt-mergebot change, but since that testing requires an example PR, I'll just merge this, make my node PR ready, and then use that to update dt-mergebot.

jakebailey:
Also, copying my thoughts on this from the discord:

> The main frustration will be that the main tsconfig (used for ts-eslint) needs to include all files, so all of those have to be included. Do you know if these packages would cause big problems if pulled into one project?
> 
> If they don't, then it's pretty simple and we can just have multiple tsconfigs with one-off test files that individually /// reference each package as a devDep
> 
> I guess the reality is that no matter what a package needs, it'll always be the case that we want some array of tsconfigs to test, so my PR is probably fine. I think the only other things to change would be running a single tsconfig only on a subset of supported versions (so, we can make tsconfigs an object, or an array of objects, or something), and the ability to share the config between tsconfigs (which I want to start supporting anyway to reduce config duplication on DT)
> 
> So, I will probably just get this in

PR #974, opened 2024-03-05
Actually print the full diff into the log when the log is too big

weswigham:
Otherwise, the "diff" we're so used to seeing reported on github hasn't actually been reported anywhere in the logs - only individual branch and main errors, not any comparisons between them.


jakebailey:
To fix the CI just do `pnpm changeset --empty`; this doesn't need a version bump as the pipeline builds from source.

packages/dtslint-runner/src/post-results.ts
```diff
@@ -46,7 +46,12 @@ async function main() {
       if (diffComment) {
         newComment += `There were interesting changes:\n`;
         if (newComment.length + diffComment.length + checkLogsMessage.length > 65535) {
-          newComment += `Changes are too big to display here, please check the log.`;
+          // hardlink directly into the output of this script
+          const detailedLogUrl = `https://typescript.visualstudio.com/TypeScript/_build/results?buildId=${buildId}&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=40b1ee41-44d6-5bba-aa04-4b76a5c732e5`;
```
jakebailey, line 50: What are the extra query params?

weswigham: Deep linking directly to the job and task within the job that refers to this script's output.

jakebailey, line 50: What are the extra query params?

jakebailey: Hm, do those still work if the pipeline changes (I want to move it to yaml at some point)

weswigham: The GUIDs will probably change if the underlying tasks shift.

weswigham: Notably, if the GUIDs don't exist in the job, it just links to the first job/task - it doesn't 404 or anything - so it's pretty safe to just use these for now and change them in the future when the underlying task changes.

jakebailey: Ah, that's not too bad then.

jakebailey (approved):

PR #975, opened 2024-03-07
Update for TS 5.4

jakebailey:
Notably we forgot to add TS 5.5 when `-next` was flipped up to it. Oops.

### DIFF
--- a/packages/typescript-versions/src/index.ts
+++ b/packages/typescript-versions/src/index.ts
@@ -6,7 +6,7 @@ import assert from "assert";
 
   For the RC:
 
-  1. Add a new version to the end of `TypeScriptVersion` and `supported`.
+  1. Add a new version to the end of `supported`.
     `supported` now contains the shipped versions, the RC, and the nightly.
   2. Add the new version to `packages/typescript-packages/package.json`.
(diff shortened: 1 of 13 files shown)


packages/typescript-versions/src/index.ts
```diff
@@ -42,9 +41,9 @@ export type AllTypeScriptVersion = UnsupportedTypeScriptVersion | TypeScriptVers
 
 export namespace TypeScriptVersion {
   /** Add to this list when a version actually ships.  */
-  export const shipped = ["4.6", "4.7", "4.8", "4.9", "5.0", "5.1", "5.2", "5.3"] as const;
+  export const shipped = ["4.7", "4.8", "4.9", "5.0", "5.1", "5.2", "5.3", "5.4"] as const;
   /** Add to this list when a version is available as typescript@next */
```
jakebailey, line 45: We didn't do when 5.5 came out, oops.

jakebailey:
Ran all of DT and it all passes, except verror which is failing in DT at 4.6/4.7, but in this PR just on 4.7.

We're not catching that in CI because our nightly runs only test `next`, which we still haven't fixed.

PR #982, opened 2024-03-22
Allow the jsxImportSource compiler option

remcohaszing:
Some JSX frameworks don’t support the classic runtime. In order to support an automatic runtime other than `react`, they must specify the option `jsxImportSource`.

See https://github.com/DefinitelyTyped/DefinitelyTyped/pull/68955#issuecomment-2014831342

### DIFF
--- a/.changeset/chilly-terms-rhyme.md
+++ b/.changeset/chilly-terms-rhyme.md
@@ -1,0 +1,5 @@
+---
+"@definitelytyped/dtslint": patch
+---
+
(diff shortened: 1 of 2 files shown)


jakebailey:
This seems good to me, but will need a changeset.

remcohaszing:
Now that I’m thinking about it, this would allow to test both the classic and the automatic runtime for `@types/react`. The tests don’t even have to change.

jakebailey (approved):

PR #986, opened 2024-04-01
Add `mime` to the list of allowed dependencies

codershiba:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69231

### DIFF
--- a/packages/definitions-parser/allowedPackageJsonDependencies.txt
+++ b/packages/definitions-parser/allowedPackageJsonDependencies.txt
@@ -749,6 +749,7 @@ meshoptimizer
 metascraper
 meteor-typings
 middy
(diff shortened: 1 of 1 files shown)


jakebailey:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69231 doesn't explain why this dep is needed.

codershiba:
> [DefinitelyTyped/DefinitelyTyped#69231](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69231) doesn't explain why this dep is needed.

Mime now ships its own types, so it will be replace `@types/mime` in that PR.

jakebailey:
This should no longer be needed.

PR #1003, opened 2024-05-15
Configure changesets to ignore private packages

jakebailey:
I didn't know about this option, but this config prevents changesets from doing anything with private packages.

Before:

image

After:

image

No publisher or retag.


jakebailey:
Uh, hm, the checker still complains. @Andarist do you know if there's a way to make changesets just never consider private packages for anything, ever? Wanting this for here and also https://github.com/microsoft/TypeScript-Website/pull/3079

Andarist:
@jakebailey I think this should address your issue: https://github.com/changesets/changesets/pull/1354 . I need to add some tests before landing this though

jakebailey:
With https://github.com/changesets/changesets/pull/1354 in, this is now working.

sandersn (approved):

PR #1021, opened 2024-06-26
Support ts-expect-error for specific version ranges

jakebailey:
This enables writing:

```ts
// @ts-expect-error >=5.5
```

Which means "only expect an error in 5.5 and above".

### DIFF
--- a/packages/eslint-plugin/src/rules/expect.ts
+++ b/packages/eslint-plugin/src/rules/expect.ts
@@ -273,7 +273,9 @@ function walk(
             if (!semver.satisfies(versionName, match[1].trim(), { loose: true })) {
               continue;
             }
-          } catch {}
+          } catch {
+            // Ignore any parsing errors.
+          }
         }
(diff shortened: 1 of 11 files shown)


andrewbranch (approved):
Awesome, thank you!

packages/eslint-plugin/src/rules/expect.ts
```diff
@@ -241,6 +242,10 @@ const zeroSourceLocation: Readonly<TSESTree.SourceLocation> = {
 
 const expectTypeToken = "$ExpectType";
 
+// Based on scanner.ts
```
sandersn, line 245: for future not-us readers:
```suggestion
// Based on Typescript's scanner.ts
```

sandersn (approved):

jakebailey:
My test isn't yet right; having a hard time getting an error that appears in 5.5 but not 5.4.

PR #1023, opened 2024-06-27
Add mergebot

sandersn:
Initial commit only so far. I made a list of tasks that I think need to happen and I'll ask for help vetting it.

- [x] Update .knip settings with the actual entrypoint(s).
- [x] Make sure name change from dt-mergebot to mergebot happens inside the copied files.
- [x] copy the deploy on push PR (and update paths now that it's in dt-tools)
- [x] copy the daily PR sync workflow
- [x] make sure workflow auth tokens are available.
- [x] copy the test workflow
- [x] make sure lint uses the repo's lint and format
- [ ] update readme on dt-mergebot to point to this location (like dts-gen, dts-critic, etc)

Later:
- [ ] Update mergebot tsconfig to use ../tsconfig-base
- [ ] Set up base *test* tsconfig.
- [ ] update dependencies
- [ ] figure out generated intermediate types, maybe by defining fragments?


packages/mergebot/package.json
```diff
@@ -0,0 +1,83 @@
...
+    "test": "jest"
+  },
+  "jest": {
+    "preset": "ts-jest",
+    "testPathIgnorePatterns": [
+      "/node_modules/",
+      "/dist/"
+    ]
+  },
+  "packageManager": "npm@10.7.0+sha256.f443ed4364ea11ac5cf7cae7fb4731278c64dd6839093f8a46eabde0430e0fcd"
```
jakebailey, line 82: You probably want to re-copy this; all of the stuff here is out of date because it uses npm

packages/mergebot/package.json
```diff
@@ -0,0 +1,83 @@
...
+  },
+  "license": "MIT",
+  "engines": {
+    "node": ">=18"
+  },
+  "dependencies": {
+    "@apollo/client": "^3.10.3",
+    "@azure/functions": "^4.4.0",
+    "@definitelytyped/old-header-parser": "npm:@definitelytyped/header-parser@0.0.178",
+    "@definitelytyped/utils": "^0.1.6",
```
jakebailey, line 23:
```suggestion
    "@definitelytyped/utils": "workspace:*",
```

packages/mergebot/package.json
```diff
@@ -0,0 +1,83 @@
...
+    "url": "https://github.com/DefinitelyTyped/dt-mergebot/issues"
+  },
+  "license": "MIT",
+  "engines": {
+    "node": ">=18"
+  },
+  "dependencies": {
+    "@apollo/client": "^3.10.3",
+    "@azure/functions": "^4.4.0",
+    "@definitelytyped/old-header-parser": "npm:@definitelytyped/header-parser@0.0.178",
```
jakebailey, line 22: Once the package-lock is gone and the root pnpm lock is updated, will need to verify that this still hits npm for the old revision.

packages/mergebot/pnpm-lock.yaml
jakebailey, line 1: This file should be deleted, then go to the root and run `pnpm install` again.

packages/mergebot/package.json
```diff
@@ -79,5 +79,5 @@
       "/dist/"
     ]
   },
-  "packageManager": "npm@10.7.0+sha256.f443ed4364ea11ac5cf7cae7fb4731278c64dd6839093f8a46eabde0430e0fcd"
+  "packageManager": "pnpm@8.15.8+sha256.691fe176eea9a8a80df20e4976f3dfb44a04841ceb885638fe2a26174f81e65e"
```
jakebailey, line 82: This line should be deleted.

packages/mergebot/package.json
sandersn: done

packages/mergebot/package.json
sandersn: Seems like yes:
```
      '@definitelytyped/old-header-parser':
        specifier: npm:@definitelytyped/header-parser@0.0.178
        version: '@definitelytyped/header-parser@0.0.178'
```

packages/mergebot/package.json
```diff
@@ -20,7 +20,7 @@
     "@apollo/client": "^3.10.3",
     "@azure/functions": "^4.4.0",
     "@definitelytyped/old-header-parser": "npm:@definitelytyped/header-parser@0.0.178",
-    "@definitelytyped/utils": "^0.1.6",
+    "@definitelytyped/utils": "workspace:",
```
jakebailey, line 23:
```suggestion
    "@definitelytyped/utils": "workspace:*",
```

packages/mergebot/src/execute-pr-actions.ts
```diff
@@ -158,7 +158,7 @@ function doRestCall(call: RestMutation): Promise<void> {
     const url = `https://api.github.com/repos/DefinitelyTyped/DefinitelyTyped/${call.op}`;
     const headers = {
         "accept": "application/vnd.github.v3+json",
-        "authorization": `token ${process.env.BOT_AUTH_TOKEN}`,
```
jakebailey: I would not recommend changing any of the token names; we have these set in the function app. If anything I would rather us become consistent and just do `GITHUB_TOKEN` like most other tools.

packages/mergebot/package.json
```diff
@@ -1,16 +1,17 @@
...
+  "version": "0.1.12",
   "description": "The bot that manages DefinitelyTyped PRs.",
   "author": "DT Maintainers",
-  "homepage": "https://github.com/DefinitelyTyped/dt-mergebot#readme",
+  "homepage": "https://github.com/microsoft/definitelytyped-tools#readme",
   "main": "dist/functions/index.js",
   "repository": {
     "type": "git",
-    "url": "https://github.com/DefinitelyTyped/dt-mergebot.git"
+    "url": "https://github.com/microsoft/definitelytyped-tools.git"
```
jakebailey, line 11: This will also need a dir set (I would check out another package)

.github/workflows/test.yml
```diff
@@ -0,0 +1,39 @@
+name: CI
```
jakebailey, line 1: Long term I don't think we need this; we already have a workflow which does testing and should be doing these tests.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,81 @@
...
+          name: node-app
+          path: ${{ env.FUNCTION_ZIP_NAME }}
+
+  deploy:
+    runs-on: ubuntu-latest
+    needs: build
+    environment:
+      name: 'Production'
+    permissions:
+      id-token: write
```
jakebailey, line 60:
```suggestion
      contents: read
      id-token: write
```

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,81 @@
...
+          pnpm run --if-present test
+      - name: Bundle
+        working-directory: packages/mergebot
+        run: pnpm run bundle
+
+      - name: Check bundle
+        run: TYPESCRIPT_BOT_TOKEN=secret node packages/mergebot/dist/functions/index.js
+
+      - name: Create zip
+        run: zip -r ${{ env.FUNCTION_ZIP_NAME }} packagaes/mergebot/dist packages/mergebot/host.json packages/mergebot/package.json
```
jakebailey, line 46: Will need to check this; the put into the zip may now be nested because we're using longer paths as the input. Better would be to consistently use `working-directory`, then output to `${{ env.GITHUB_WORKSPACE }}/${{ env.FUNCTION_ZIP_NAME }}`.

packages/mergebot/src/execute-pr-actions.ts
sandersn: oh, right. I got confused between the variables available to workflows vs functions.

jakebailey (approved):

PR #1024, opened 2024-06-28
Clear expect rule program cache when heap is almost exhausted

andrewbranch:
There is some controversy over whether this is completely reliable, but it seems to work well in my tests, with and without `--max-old-space-size` passed in.

I initially tried to just stop adding new programs to the cache at a certain threshold, but somehow memory still shoots up after no more programs are being created at a modest heap usage. So we do this check and purge even if the current `getProgram` call didn’t eat up more memory itself.

This fixes the recent openui5 CI OOM.

### DIFF
--- a/packages/eslint-plugin/src/rules/expect.ts
+++ b/packages/eslint-plugin/src/rules/expect.ts
@@ -1,10 +1,11 @@
-import { createRule, findTypesPackage, findUp } from "../util";
 import { ESLintUtils, TSESTree } from "@typescript-eslint/utils";
-import type * as ts from "typescript";
-import path from "path";
-import fs from "fs";
 import { ReportDescriptorMessageData } from "@typescript-eslint/utils/ts-eslint";
+import fs from "fs";
+import v8 from "node:v8";
+import path from "path";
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin/src/rules/expect.ts
```diff
@@ -198,6 +199,12 @@ function getProgram(
     newProgram = createProgram(path.resolve(dirPath, configFile), ts);
     versionToProgram.set(cacheKey, newProgram);
   }
+
+  const heapUsage = v8.getHeapStatistics().used_heap_size / v8.getHeapStatistics().heap_size_limit;
```
jakebailey, line 203:
```suggestion
  const heapStats = v8.getHeapStatistics();
  const heapUsage = heapStats.used_heap_size / heapStats.heap_size_limit;
```

jakebailey (approved):
Needs a changeset but otherwise seems good.

jakebailey:
Though, hopefully it won't get into a situation where GC hasn't run and the next run throws things away again. I'd say to call `process.gc` but that won't be available without `--expose-gc`...

PR #1028, opened 2024-07-03
remove array-type eslint rule

iisaduan:
The `"array-type":"error"` behavior, seems to have changed in the most recent typescript-eslint release, and as a result, it is generating new [errors](https://github.com/DefinitelyTyped/DefinitelyTyped/actions/runs/9777487309). 

With the addition of the default setting, the error message also becomes wrong:
image


jakebailey:
Isn't this error right? Readonly of array is not ReadonlyArray and that's usually always a mistake...

iisaduan:
>Isn't this error right? Readonly of array is not ReadonlyArray and that's usually always a mistake...

The error message changes when you remove the `default`, so at least one of the messages are wrong

image

iisaduan:
https://github.com/typescript-eslint/typescript-eslint/issues/9498 well I filed a bug

jakebailey (commented):
Removing it seems fine, but this PR needs a changeset.

sandersn (approved):

jakebailey:
Updated the PR so we can get it in.

jakebailey (approved):

PR #1031, opened 2024-07-09
Switch mergebot to new project boards

sandersn:
New new project board is at https://github.com/orgs/DefinitelyTyped/projects/1 (better name coming soon).

Things the new project board automation could probably replace:

- Adding new PRs to the board, probably under Other. The graphql API doesn't let you add cards to anything but a non-status "Not Started", and then requires a follow-up move.
- Deleting closed PRs from the board. Right now we have a 4-times-a-day run that does this.

I think I'll address those in followup PRs.


packages/mergebot/src/scripts/updateJSONFixtures.mjs
```diff
@@ -0,0 +1,50 @@
+// @ts-check
```
jakebailey, line 1: This is now mjs; was this script never referenecd elsewhere? Or is this a new thing entirely form the old one? Is there a doc that needs to be updated?

jakebailey: Oh, the old one was some travis thing :smile:

sandersn: It used to be updateJSONFixtures.ts, and run with ts-node. It's not referenced anywhere else, and not documented.

jakebailey (approved):
What's here looks good to me.

packages/mergebot/src/scripts/updateJSONFixtures.mjs
sandersn: There's no reason for it to be .ts except to use module syntax, and it's way easier to run (with node >=20) as a JS file.

packages/mergebot/src/run.ts
```diff
@@ -26,6 +26,7 @@ const args = yargs(process.argv.slice(2))
     "show-extended": { alias: ["s3"], type: "boolean", desc: "display extended info" },
     "show-actions": { alias: ["s4"], type: "boolean", desc: "display actions" },
     "show-mutations": { alias: ["s5"], type: "boolean", desc: "display mutations" },
+    "project-only": { type: "boolean", desc: "Only change the project board, nothing else." },
```
sandersn, line 29: project-only is useful for local runs but not much else, so I don't know if it's worth keeping. I left it in because maybe it'll be useful elsewhere.

packages/mergebot/src/run.ts
```diff
@@ -117,48 +119,43 @@ const start = async function () {
...
   // Handle other columns
-  for (const column of columns) {
-    if (column.name === "Recently Merged") continue;
-    const ids = column.cards.map((c) => c.id).filter((c) => !cardIDs.includes(c));
+  for (const [name, cards] of columns) {
+    if (name === "Recently Merged") continue;
+    const ids = cards.map((c) => c.id).filter((c) => !cardIDs.includes(c));
     if (ids.length === 0) continue;
-    console.log(`Cleaning up closed PRs in "${column.name}"`);
-    // don't actually do the deletions, until I follow this and make sure that it's working fine
```
sandersn, line 158: this now actually does the deletions; I checked locally that it's working fine

sandersn (commented):
Some comments before I merge.

PR #1032, opened 2024-07-11
New project board: 2 small followups

sandersn:
1. Update the webhook for manual card moves. The code compiles but I have no idea how to test it -- I don't know what the rules are for manual moves that the bot will respect.
2. Fix the "Should delete" output of run.ts to not print for open PRs, only closed and merged ones that were created, then closed, during the run of run.ts itself. Also add "Deleting" output when *actually* deleting a PR.
3. Fix one now-incorrect syntax for a `pnpm` call in the README (`--` is no longer needed).

### DIFF
--- a/packages/mergebot/README.md
+++ b/packages/mergebot/README.md
@@ -99,7 +99,7 @@ To create fixtures of a current PR:
 
 ```sh
 # To create a fixture for PR 43161
-pnpm run create-fixture -- 43161
+pnpm run create-fixture 43161
 ```
 
 Then you can work against these fixtures offline with:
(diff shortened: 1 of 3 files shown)


jakebailey (approved):
I feel like this is correct but I don't think I've ever blessed PRs via the board myself, so can't really test it. Odd we have no tests but I guess we'd have needed to capture PR right when it was moved on the board but had not yet been reprocessed...

sandersn:
I'm going to merge and watch the logs while I move a card to make sure that it at least doesn't crash while handling the event.

PR #1033, opened 2024-07-12
Load source maps in mergebot

jakebailey:
To aid in debugging crashes.

### DIFF
--- a/packages/mergebot/package.json
+++ b/packages/mergebot/package.json
@@ -28,6 +28,7 @@
     "fast-json-patch": "^3.1.1",
     "graphql": "15.8.0",
     "prettyjson": "^1.2.5",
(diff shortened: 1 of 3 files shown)


jakebailey:
Alternatively we can set `languageWorkers__node__arguments=--enable-source-maps` on Azure.

jakebailey:
I tried setting:

```
--source-root=https://raw.githubusercontent.com/microsoft/DefinitelyTyped-tools/main/packages/mergebot/dist
```

Per https://esbuild.github.io/api/#source-root, but that doesn't work with `source-map-support`, only with Node's `--enable-source-maps`. Hmmm.

sandersn (approved):

PR #1034, opened 2024-07-12
Query board fields instead of cards for columns

jakebailey:
This should fix merges, however, the test passed even when it was wrong, Only `create-fixture` would have shown the error...


packages/mergebot/src/queries/schema/CardIdToPr.ts
```diff
@@ -10,249 +10,7 @@ import { PullRequestState } from "./graphql-global-types";
 // ====================================================
 
 export interface CardIdToPr_node_AddedToMergeQueueEvent {
-  __typename:
```
jakebailey, line 13: Fix this fmt thing, prettier wasn't ignoring these generated files.

jakebailey:
I had to update the way that we did caching of the data; the old one wasn't updated for the projects v2 and had the side effect of only working because it wasn't run (which was why tests behaved weird).

sandersn:
yeah, I think I missed caching entirely

PR #1042, opened 2024-08-07
Refactor suspicious config checking to be strongly typed

jakebailey:
A little overkill but I wanted to be sure.


jakebailey:
Hm, the crash went away, even though I ran it after the merge...

sandersn:
To me, it looks like there were no failing runs after the merge of the `delete data.compilerOptions?.[k]` PR. Maybe I'm misreading something.

(For the first passing one, I double checked that 70218 was still open -- it was)

sandersn (approved):
I think this is an improvement if you think it's a good idea to change the code at all.

PR #1043, opened 2024-08-07
Keep the newest 50 merged PRs, not the oldest 50 merged PRs

jakebailey:
I've been wondering why ancient PRs are on the board. The code here sorts the dates _ascending_, then throws away the first 50, deleting the rest. Except, that's the newest entries, which are the most important ones. Oops.


sandersn (approved):
very good bug

PR #1045, opened 2024-08-07
Fix blessings

jakebailey:
This PR:

- Removes the "blessed" state when in "Needs Merged", which allowed authors to merge without an approval. A maintainer can just click "approve". That's already the same thing.
- Creates a new "Waiting for Code Reviews (Blessed)" column; moving a card into this column allows the PR to bypass maintainer checks, but still requires owner approval. This is the same as the old blessing.

I left the old code behind, which we can reuse if/when GitHub adds the API to get this info,


andrewbranch (approved):

packages/mergebot/src/basic.ts
```diff
@@ -6,8 +6,22 @@ export type ColumnName =
...
   | "Waiting for Code Reviews"
+  | BlessedColumnName
   | "*REMOVE*"; // special value: indicates closing the PR
 
+export type BlessedColumnName = (typeof blessedColumnNames)[number];
+const blessedColumnNames = [
+  "Waiting for Code Reviews (Blessed)",
+] as const;
+
+export function isBlessedColumnName(name: string): name is BlessedColumnName {
```
sandersn, line 17: are we ever going back to multiple blessed columns? This seems like overkill otherwise. for example, "Waiting for Code Review (Blessed)" is hard-coded all over compute-pr-actions.ts but not pr-info.ts. Why not both?

packages/mergebot/src/basic.ts
```diff
@@ -50,4 +64,17 @@ export const labelNames = [
 
 export type ApproverKind = "maintainer" | "owner" | "other";
 
-export type BlessingKind = "merge" | "review" | undefined;
+// https://github.com/orgs/DefinitelyTyped/projects/1
+export const projectBoardNumber = 1;
```
sandersn, line 68: 👍🏼 I forgot about this file when I added these constants.

sandersn (approved):
Looks fine, although I'd personally have `type BlessedColumnName = "Waiting ...` and remove blessedColumnNames, isBlessededColumn, and maybe columnNameToBeBlessed, inlining the check against the single column name.

packages/mergebot/src/basic.ts
jakebailey: Probably not, but an earlier revision of the PR did. I can simplify this.

PR #1046, opened 2024-08-08
Move blessed mergable PRs into new column, fix blessed PRs with partial approvals

jakebailey:
~The new test reveals that blessing also bypasses needing _all_ owners of modified packages to approve. Working on figuring that one out.~


packages/mergebot/src/basic.ts
```diff
@@ -9,7 +9,21 @@ export type ColumnName =
   | BlessedColumnName
   | "*REMOVE*"; // special value: indicates closing the PR
 
-export type BlessedColumnName = "Waiting for Code Reviews (Blessed)";
+export type BlessedColumnName = (typeof blessedColumnNames)[number];
+
+const blessedColumnNames = [
+  "Waiting for Code Reviews (Blessed)",
+  "Waiting for Author to Merge (Blessed)",
```
sandersn, line 16: does the weekly maintainer have to know which (Blessed) column to move to?

sandersn (approved):
Code looks OK but I don't think I know how to *use* this feature.

packages/mergebot/src/basic.ts
jakebailey: No, they should only ever put something into the first one, since I removed the "straight to mergable" thing.

jakebailey:
The PR's not yet complete; working on a fix for the blessed PRs overriding any other reviews thing.

packages/mergebot/src/compute-pr-actions.ts
```diff
@@ -114,9 +114,7 @@ function extendPrInfo(info: PrInfo): ExtendedPrInfo {
   const changereqReviews = info.reviews.filter((r) => r.type === "changereq") as ExtendedPrInfo["changereqReviews"];
   const staleReviews = info.reviews.filter((r) => r.type === "stale") as ExtendedPrInfo["staleReviews"];
   const hasChangereqs = changereqReviews.length > 0;
-  const approvedBy = getApprovedBy();
-  const pendingCriticalPackages = getPendingCriticalPackages();
-  const approverKind = getApproverKind();
```
jakebailey, line 119: Dropped some of these to make them less of a hazard for code below; a lot of code was incorrectly reading out reviews for one package and applying them to the whole.

jakebailey:
> I don't think I know how to _use_ this feature.

The use is just: If there's a PR in "needs maintainer review" and you think it's not doing anything harmful such that an owner can make the final decision, move the PR to "Waiting for Code Reviews (blessed)".

If all authors approve, it will eventually move to "Waiting for Merge (Blessed)" and be eligible for a self merge. If someone at any point complains, it leaves the "blessed" part of the flow, dropping back into "Needs Author Action", or eventually back into the maintainer queue.

packages/mergebot/src/basic.ts
```diff
@@ -53,6 +67,12 @@ export const labelNames = [
 
 export type ApproverKind = "maintainer" | "owner" | "other";
```
sandersn, line 68: stupid idea (because the code is already good enough): why not `export type ApproverKind = (typeof approverKindOrder)[number]` ?

but also: numeric enums give you all the features you need with `enum ApproverKind { None, Other, Owner, Maintainer }` to simplify `getMaxApproverKind`.

sandersn (approved):
Looks good. One very optional suggestion about enums.

packages/mergebot/src/basic.ts
jakebailey: I'll do the former one, but I was avoiding the latter because the ApproverKind is a part of baselines and will end up showing up in the dumped JSON as regular numbers.

PR #1054, opened 2024-09-13
fix: Remove async-retry from expectedNpmVersionFailures.txt

hkleungai:
fix: Remove async-retry from expectedNpmVersionFailures.txt

Upon successful merge of [DefinitelyTyped PR #70552](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/70552), `async-retry` can be removed from `expectedNpmVersionFailures.txt`.


jakebailey:
The linked PR means releasing a lower version; I'm not sure that leads to a good result (I commented there)

Issue #1056, opened 2024-09-18
types-registry stopped publishing?


moltar:
Last release was 4 months ago?

https://www.npmjs.com/package/types-registry?activeTab=versions

jakebailey:
Yeah, failing: https://github.com/microsoft/DefinitelyTyped-tools/actions/workflows/publish-registry.yml

jakebailey:
0.1.686 was just published successfully. Thanks for telling us; need to add some badges to the readme or something...

PR #1057, opened 2024-09-18
Attempt to fix types-registry with retry / nAtATime

jakebailey:
Trying to fix https://github.com/microsoft/DefinitelyTyped-tools/issues/1056

### DIFF
--- a/packages/publisher/src/publish-registry.ts
+++ b/packages/publisher/src/publish-registry.ts
@@ -29,6 +29,8 @@ import {
   NpmPublishClient,
   cacheDir,
   isObject,
+  nAtATime,
+  retry,
 } from "@definitelytyped/utils";
 import * as pacote from "pacote";
 import * as semver from "semver";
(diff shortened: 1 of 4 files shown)


packages/publisher/src/publish-registry.ts
```diff
@@ -233,11 +235,20 @@ interface Registry {
 async function generateRegistry(typings: readonly TypingsData[]): Promise<Registry> {
   return {
     entries: Object.fromEntries(
-      await Promise.all(
-        typings.map(async (typing) => [
-          typing.typesDirectoryName,
-          filterTags((await pacote.packument(typing.name, { cache: cacheDir }))["dist-tags"]),
-        ]),
```
jakebailey, line 240: Might be wrong but I'm pretty sure this was attempting to request all 8000 some packuments at once?

sandersn: seems like it

sandersn (approved):

PR #1067, opened 2024-09-26
Adding axios-mock-adapter to allowed packages list

camilourd:


camilourd:
Adding axios-mock-adapter to allowed packages list

jakebailey:
What PR requires this change?

camilourd:
The second

jakebailey:
That doesn't answer my question... But I assume it's for https://github.com/DefinitelyTyped/DefinitelyTyped/pull/70695?

jakebailey:
As noted on https://github.com/DefinitelyTyped/DefinitelyTyped/pull/70695#pullrequestreview-2338115612, you own the package you're trying to add types for. You can already ship types yourself in your own package.

camilourd:
yes, but it's failing the test due to a library called axios-mock-adaptor. It suggested to do this pull request to add this library to the available list, but I'm noticing that I will have to change it to an Demo type to avoid future problems

jakebailey:
Again, you own the package; you were able to send a PR from the same github username as the package links to for its source. The package also already contains `d.ts` files: https://unpkg.com/browse/kyc-tools@1.0.9/dist/index.d.ts

There's no need for either PR.

jakebailey (changes requested):
This is not needed, per above

PR #1071, opened 2024-09-29
feat: Add esbuild to allowedPackageJsonDependencies.txt

hkleungai:
There are existing esbuild-plugin in the DT repo currently, including one of my ongoing [PR #70733](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/70733). To facilitate adding or modifying these types in DT side, I would like to ask DT-tools maintainer to kindly accept and merge this PR.

### DIFF
--- a/packages/definitions-parser/allowedPackageJsonDependencies.txt
+++ b/packages/definitions-parser/allowedPackageJsonDependencies.txt
@@ -718,6 +718,7 @@ ember-qunit
 ember-resolver
 ember-source
 eris
+esbuild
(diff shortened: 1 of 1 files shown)


jakebailey:
This would be a pretty unfortunate dep to add in the repo thanks to just how many platform-specific deps this will pull in; is there no way to declare things such that this dependency is not needed?

jakebailey (approved):
On second thought, this is fine. We allow react-native, which is many times worse than this.

PR #1072, opened 2024-09-30
Add support for peer deps

### LINKED ISSUE #433: Allow peer dependencies with stricter version ranges
Continuation of https://github.com/microsoft/types-publisher/issues/431.

## Current state

When you import from a package hosted in the DefinitelyTyped repository in a package that's also hosted in DefinitelyTyped, the types publisher will automatically declare a direct `dependency` on that types package matching any version. For example, if you have a package `react-library` that imports from `react` then the `package.json` of `@types/react-library` will look something like this:
```json
{
  "name": "@types/react-library",
  "dependencies": {
    "@types/react": "*"
  }
}
```

## Problems

Combination of the described problems are encountered in
- https://github.com/facebook/react/issues/24304
- https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/59839
- https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/59820
- https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/59841

### duplication of `@types/*` packages

Given an existing module graph similar to 
```bash
types-peer-deps@1.0.0 /home/eps1lon/Development/throwaway/types-peer-deps
├─┬ @types/react-library@1.0.0
│ └── @types/react@17.0.44 deduped
└── @types/react@17.0.44
```
If you want to upgrade to React 18 you would need to update `@types/react` with `npm install @types/react@18`.
However, `npm` (and `yarn` and probably other package managers) make sure that this install does not alter the dependency tree too much (if you're using lockfiles). What you end up with is something like
```bash
types-peer-deps@1.0.0 /home/eps1lon/Development/throwaway/types-peer-deps
├─┬ @types/react-library@1.0.0
│ └── @types/react@17.0.44 
└── @types/react@18.0.0
```

Now your dependency tree contains multiple version of a types package. In this specific case these packages are incompatible.
To fix, one would have to apply one of the workarounds described in https://github.com/facebook/react/issues/24304#issuecomment-1094565891 which isn't trivial at all.
After all, package managers are supposed to resolve these issues. But they can't if we provide them with inaccurate information.

### `*` is not forwards compatible

While currently published packages might work perfectly fine with all published versions of `@types/react` at the time, there's no guarantee that continues to work. It essentially requires that no breaking changes are made at all which defeats the purpose of SemVer major releases.

### Allow peer dependencies

React is generally a peer dependency and therefore it types should as well.
If we could move `@types/react` from `dependencies` to peer dependencies with
```json
"peerDependencies": { 
  "@types/react": "*"
},
```
we would ensure types deduplication since the package owner would defines the version of `@types/react` a single time (via `dependencies` or `devDependency`.

### Allow constraining `@types/*` dependencies to something stricter than `*`

Either let maintainers decide the range manually or infer the range from the versions it's currently tested on. For example, `@types/react-library` runs tests with `@types/react` v15, v16, v17 and latest. In that case we check what version the current release has (or will be) and declare that. 
This way an update to `@types/react-library` will also bump versions of `dependencies` and therefore propagate bug fixes and features of those dependencies upon release.

## related issues

- https://github.com/DefinitelyTyped/DefinitelyTyped/issues/33015
- https://github.com/DefinitelyTyped/DefinitelyTyped/issues/56011
- https://github.com/microsoft/types-publisher/issues/431

jakebailey:
Fixes #433

We've hit multiple cases now where peer deps would have really been beneficial, and now with the new (well, not new anymore) monorepo layout, we can handle these without much code change.

Likely needs more tests, though there really isn't much here besides plumbing.


jakebailey:
It's worth nothing that peer deps are intended to be used in cases where the user is _already_ going to install that peer dep. E.g. publishing a react component, where you peer dep on `@types/react` because the user will have already installed it and can control the version.

I am not sure how we can possibly enforce that except via human interaction. Perhaps we should make `mergebot` flag any PR which modifies `peerDeps` as needing a maintainer review?

sandersn:
I think maintainer review is a good first state for any confusing systemic change. We can remove the requirement if it's onerous (but I doubt it'll be common, so I bet it's not a problem.)

jakebailey:
Added a mergebot complaint which should put `Check Config` on those PRs, though I don't have a PR to test it on quite yet. Will make one at the risk of annoying someone...

packages/publisher/test/generate-packages.test.ts
```diff
@@ -114,6 +116,13 @@ testo({
       ),
     );
   },
+  readmeOnePeer() {
+    const dt = defaultFS();
+    const typing = new TypingsData(dt.fs, createRawPackage(License.Apache20), /*isLatest*/ true);
+    expect(createReadme(typing, dt.pkgFS("jquery"))).toEqual(
+      expect.stringContaining("Peer dependencies: [@types/express](https://npmjs.com/package/@types/express)"),
```
sandersn, line 123: is this the most obvious expect syntax for `stringContaining`? I'd *expect* something like `expect(createReadme(...)).stringContaining("...")`

The two things that weird me out are
1. toEqual, but then it's not actually testing full equality.
2. expect.stringContaining, which seems like namespacing, when all the other uses of `expect` are callable, meant to be read like a sentence.

This probably *is* the most direct way to write this, and `expect` syntax is just intentionally awful.

sandersn (approved):
Looks good so far. But I don't have a way to tell from a review whether it's a complete change. I guess the mergebot snapshot test will help with that.

packages/publisher/test/generate-packages.test.ts
jakebailey: Probably it'd be better to snapshot test this

jakebailey: (I just copied from the test above)

jakebailey:
I went ahead and added proper tests for the disallowed dep errors.

packages/publisher/test/generate-packages.test.ts
jakebailey: Yeah, all of the tests are like this, it's just how it's done: https://jestjs.io/docs/expect#expectstringcontainingstring

Weird...

jakebailey:
@zkochan Do you have any thoughts on whether or not `@types` packages gaining peer deps will cause a problem for pnpm? I'm not sure what special casing `pnpm` does for types packages at the moment, but having these is likely to improve resolution situations with things like react components.

zkochan:
pnpm struggles a lot with peer dependencies. The issue is that peer dependencies are used too frequently nowadays. To make things worth peer dependencies have peer dependencies of their own and some peer dependencies even form a circular connection. I am not sure how other package managers solve these issues. Maybe it is less of an issue in a hoisted node_modules. pnpm, however, needs to consciously resolve all the peer dependency combinations and because of the above complexities it becomes a very expensive operation. I've spent many hundreds of hours optimizing the peer dependencies resolution part of pnpm.

So, if this change will result in even more peer dependencies, pnpm install will use more memory during peer dependencies resolution. Hopefully it won't cause out-of-memory errors.

Maybe if we could kind of group the type peer with the package peer that it is related to and resolve them as one entity, then it wouldn't make peers resolution a harder task.

jakebailey:
My expectation is that basically all `@types/react-*` ish packages will end up with peer deps on `@types/react`, and then some other examples similar to that. The intent is that types packages don't use peer deps unless the package they're describing have peer deps (or, _should_ have declared something as a peer dep).

Hopefully the result is not too much more memory ☹️

PR #1075, opened 2024-10-09
feat: Remove leaflet-gpx from expectedNpmVersionFailures.txt

tbdrz:

PR #1080, opened 2024-10-14
Fix missing changeset CI PR check

jakebailey:
This workflow was checking `origin/main` rather than the merge base of the PR; due to race conditions or other timing issues with manual CI run approval, we may end up in a situation where the two are different, leading to other unrelated changes showing up in the diff during the changeset check. (Like https://github.com/microsoft/DefinitelyTyped-tools/actions/runs/11257648935/job/31524626555?pr=1075)


jakebailey:
Merged to unblock #1075, which it does.

PR #1082, opened 2024-10-17
Update to typescript-eslint v8

jakebailey:
`ban-types` is out, `no-unsafe-function-type` and `no-wrapper-object-types` are in.

DT will have to be updated to deal with this.


andrewbranch (approved):

jakebailey:
I have absolutely no idea why this is failing in CI and not locally :(

andrewbranch:
I recall a previous PR of mine where that same rule was failing _locally_ but not in CI. I don’t think I ended up figuring it out 😕

packages/eslint-plugin/src/rules/no-type-only-packages.ts
```diff
@@ -32,7 +32,11 @@ const rule = createRule({
     let hasNonEmptyFile = false;
 
     function isOutsidePackage(fileName: string): boolean {
-      return fileName.includes("node_modules") || path.relative(pkg!.dir, fileName).startsWith("..");
+      if (fileName.includes("node_modules") || path.relative(pkg!.dir, fileName).startsWith("..")) {
+        return true;
+      }
+      const other = findTypesPackage(fileName);
+      return !other || other.dir !== pkg!.dir;
```
jakebailey, line 39: It turns out that this test explicitly includes files from other packages (including versioned subdirs) to test the rule, but we weren't actually checking this here. It's not clear to me why this wasn't breaking before, though.

jakebailey: It's technically not the best to not cache this result, but since linting can be in an editor, we don't want to store a global cache or something. It shouldn't really cost that much.

jakebailey:
This is all working now, though DT itself needs to be fixed to deal with the ban-types changes, so that will have to come later.

sandersn (approved):
Ping me on the DT review when it's ready.

jakebailey:
I have no idea how but this PR broke publishing; it's repeatedly publishing deprecated packages that it's already deprecated. I've paused it until I can figure out what the heck is happening.

PR #1120, opened 2024-12-09
Fix incorrect newlines in package readmes with peer deps

jakebailey:
`react-dom` used to look like:

image

But now looks like this after adding peer deps:

image

When I added peer deps, I missed that I screwed up this formatting.


andrewbranch (approved):

jakebailey:
I'm rebasing this with some testing changes, since this was clearly not tested and still isn't.

jakebailey:
https://github.com/microsoft/DefinitelyTyped-tools/pull/1120/commits/95f8f081d9fc9f619f88952171abedb2737235d4 now shows exactly what changed.

PR #1131, opened 2025-01-12
add windows-cpu to allowedPackageJsonDependencies.txt

KontrollFreek:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/71635


jakebailey:
See https://github.com/DefinitelyTyped/DefinitelyTyped/pull/71635#discussion_r1912807421; there's no reason to add a package as a dep to its own types. The only packages that should be in this list are packages that have types that are needed on DT.

PR #1133, opened 2025-01-17
feat: add scripts to check pkg in attw / expectedVersionFailures

hkleungai:
A much more formal script setup for https://github.com/microsoft/DefinitelyTyped-tools/pull/1130. See if the code makes sense to DT-tools maintainer first, then I will proceed to incorporating them into github actions.

This PR involve some refactoring on existing dtslint code. Few things you may want to take notes on.

## feat(dtslint): apply getMangledNameForScopedPackage() for package name

`@definitelytyped/definitions-parser` is added as new dep for `getMangledNameForScopedPackage()`. This shall faciliate code reuse.

## refactor: extract functions into util folder with minor fix

Some assertions in `packages/dtslint/src/index.ts` is moved to `packages/dtslint/src/util.ts`, with small modifications.

In `findDTRootAndPackageNameFrom()`, the assertions are reordered, to make the code more readable.

In `assertPathIsInDefinitelyTyped()`, the condition is modified. Hence if a non-existing type lib is used to run test in DT, e.g. `pnpm run abc`, then the error is this assertion will be properly thrown. Note that it will replace the current behaviour of throwing `Error: ENOENT: no such file or directory, scandir '/<root-dir>/DefinitelyTyped/types/abc'` in non-existing type lib test.

## feat: add scripts to check pkg in attw / expectedVersionFailures

Two relatively simple scripts are added, aiming to check (failed) package in `attw.json` and `expectedVersionFailures.txt` in future github action development.


packages/dtslint/package.json
```diff
@@ -24,6 +24,7 @@
   "dependencies": {
     "@arethetypeswrong/cli": "0.15.1",
     "@arethetypeswrong/core": "0.15.1",
+    "@definitelytyped/definitions-parser": "workspace:*",
```
jakebailey, line 27: Make sure the tsconfig is updated to declare this dependency too.

pnpm-lock.yaml
jakebailey, line 1: This diff is weird, can you restore it to main and then just run `pnpm install` to see if it just adds the one new dep edge?

packages/dtslint/src/index.ts
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 
+import { getMangledNameForScopedPackage } from "@definitelytyped/definitions-parser";
```
jakebailey, line 3: Hm, I thought we had something like this in the utils package.

packages/dtslint/package.json
hkleungai: This package is now removed.

packages/dtslint/src/index.ts
hkleungai: Thank you for the hint. Updated.

PR #1153, opened 2025-03-27
Added redux toolkit and p2party allowed deps for DefinitelyTyped

fuzzc0re:


packages/definitions-parser/allowedPackageJsonDependencies.txt
```diff
@@ -585,6 +586,7 @@
 @types/next-redux-wrapper
 @types/node
 @types/node-fetch
+@types/p2party
```
andrewbranch, line 589: I think you don’t need this—these `@types` packages are only here if another DT package needs to depend on a non-current version of the package, i.e. they need to download it from npm instead of linking in the package from the repo.

andrewbranch (changes requested):

packages/definitions-parser/allowedPackageJsonDependencies.txt
andrewbranch: @jakebailey removed all the `@types` packages from this list. If you fix conflicts by removing this addition, we can merge.

PR #1160, opened 2025-04-02
feat: Remove use-sync-external-store from expectedNpmVersionFailures.txt

eps1lon:
For once https://github.com/DefinitelyTyped/DefinitelyTyped/pull/72360 landed


sandersn (approved):

andrewbranch:
This list means that DefinitelyTyped declares a major version of a types package that doesn’t have a matching implementation on npm. attw can’t possibly run in that scenario since there’s no implementation package version it can use as a source of truth. The DT PR bumps use-sync-external-store to a version that actually exists, so attw can start running.

PR #1169, opened 2025-04-11
mergebot: remove missed project V1 code

sandersn:
It hasn't done anything since switching to the V2 query, but now the github API returns an error when mentioning V1.


packages/mergebot/src/queries/all-open-prs-query.ts
```diff
@@ -3,18 +3,13 @@ import { client } from "../graphql-client";
 import { GetAllOpenPRsAndCardIDs, GetAllOpenPRsAndCardIDsVariables } from "./schema/GetAllOpenPRsAndCardIDs";
 import { noNullish } from "../util/util";
 
-const getAllOpenPRsAndCardIDsQuery: TypedDocumentNode<GetAllOpenPRsAndCardIDs, GetAllOpenPRsAndCardIDsVariables> = gql`
+const getAllOpenPRsQuery: TypedDocumentNode<GetAllOpenPRsAndCardIDs, GetAllOpenPRsAndCardIDsVariables> = gql`
```
jakebailey, line 6: Doesn't modifying this require committing the query types?

sandersn: In theory yes, I think so. I think it works because I'm just deleting some code.

jakebailey: Yeah, you should run `AUTH_TOKEN=$(gh auth token) pnpm run graphql-schema` in the mergebot dir. I didn't realize CI was not enforcing that this was clean.

jakebailey (approved):

PR #1175, opened 2025-04-23
Ignore drive-by reviews in mergebot

jakebailey:
I noticed that anyone can go to any PR and request changes, which makes the bot say "One or more reviewers has requested changes. Please address their comments. I'll be back once they sign off or you've pushed new commits. Thank you!" and applies the "Revision needed" tag.

I feel like this shouldn't be possible; we shouldn't care about reviews from people who can't actually approve the PR.


packages/mergebot/src/_tests/fixtures/43136/mutations.json
```diff
@@ -52,14 +52,5 @@
         "body": "🔔 @RReverser — please [review this PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/43136/files) in the next few days. Be sure to explicitly select **`Approve`** or **`Request Changes`** in the GitHub UI so I know what's going on.\n<!--typescript_bot_pinging-reviewers-->"
       }
     }
-  },
-  {
-    "mutation": "mutation ($input: AddCommentInput!) {\n  addComment(input: $input) {\n    __typename\n  }\n}\n",
-    "variables": {
-      "input": {
-        "subjectId": "MDExOlB1bGxSZXF1ZXN0Mzg4MzgyMzYy",
-        "body": "@RReverser Thank you for reviewing this PR! The author has pushed new commits since your last review. Could you take another look and submit a fresh review?\n<!--typescript_bot_stale-ping-f5a3ad-37d784d-->"
```
jakebailey, line 61: This is an example of a downside; if randos do come and review, anything they say will be ignored and they won't be pinged.

In this case, I doubt RReverser is a rando 😄

packages/mergebot/src/_tests/fixtures/43160/result.json
```diff
@@ -2,7 +2,6 @@
   "projectColumn": "Needs Author Action",
   "labels": [
     "The CI failed",
-    "Owner Approved",
```
jakebailey, line 5: This might be a bug; odd to have an owner approval get removed...

sandersn (changes requested):
With the amount of TS team oversight still spent on DT, I'd say that the current state is correct, because we still have the ability to override driveby comments with an approval or merge.

But typically any comment at all means that the PR isn't settled and shouldn't merge. I guess the exceptions are "good, glad we came to an agreement!" kinds of things, but that's less common than "wait wait I disagree" ones. I think.

PR #1180, opened 2025-05-12
Update ATTW to 0.18.1

jakebailey:
To get https://github.com/arethetypeswrong/arethetypeswrong.github.io/pull/245


andrewbranch:
This upgrade includes https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/NamedExports.md, basically the same thing that dts-critic used to surface as a suggestion. We’ll probably need to disable it in DT's attw.json in order not to get a bunch more failures.

PR #1183, opened 2025-06-23
Add eslint to allowed packages

MysteryBlokHed:
Related PR: DefinitelyTyped/DefinitelyTyped#73093

A few types from the eslint package are used to define a plugin here. The `@types/eslint` package no longer exists (removed in DefinitelyTyped/DefinitelyTyped#70735), so that can't be used.


jakebailey (approved):

jakebailey:
Oh boy, I can't wait to see what breaks when we put eslint inside of our packages, where we need eslint to lint he packages

PR #1187, opened 2025-07-10
Mark PRs with nonNpm as Check Config

jakebailey:
People still try to contribute packages which go against https://github.com/DefinitelyTyped/DefinitelyTyped#what-about-type-definitions-with-no-matching-package

Probably worth flagging `nonNpm` packages as `Check Config`.


sandersn (approved):
Check Config means that a maintainer must approve it, right?

jakebailey:
No, I don't think it has that effect. Will check

jakebailey:
It would. Nevermind.

PR #1188, opened 2025-07-29
Remove strip-ansi in favor of stripVTControlCharacters

jakebailey:
Learned something new this weekend; `stripVTControlCharacters` has been in Node since 16.11 so there's no need for `strip-ansi` anymore.


.changeset/thick-donkeys-flash.md
```diff
@@ -0,0 +1,2 @@
+---
```
sandersn, line 1: these names get weirder and weirder

sandersn (approved):

PR #1199, opened 2025-09-24
Add `@sass/types` and `sass-embedded` to allowed dependencies

stof:
In https://github.com/DefinitelyTyped/DefinitelyTyped/pull/73736, I defined the type declarations for `gulp-sass` v6.

In those type declarations, I'm using the `@sass/types` package, which is a new type-only package published by the Sass team with declarations from the official Sass JS API.
`sass-embedded` is used only as dev dependency, to ensure that `gulp-sass` types indeed work both for using `gulp-sass` with `sass` and for using it with `sass-embedded`.

### DIFF
--- a/packages/definitions-parser/allowedPackageJsonDependencies.txt
+++ b/packages/definitions-parser/allowedPackageJsonDependencies.txt
@@ -542,6 +542,7 @@
 @react-navigation/native
 @react-navigation/stack
 @redux-saga/types
+@sass/types
 @sentry/browser
 @serverless/typescript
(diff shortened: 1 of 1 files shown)


andrewbranch (approved):

stof:
@andrewbranch what should I do in my DT PR now, where the CI was failing asking me to open a PR here ? Do I need to do a no-op change in my branch just to trigger a new CI run using the updated runner ? Or is the process different ?

jakebailey:
A close reopen will do it

PR #1202, opened 2025-09-28
Add ipaddr.js to allowed dependencies

ultraslayyy:


ultraslayyy:
Oh sorry forgot to mention.
In DefinitelyTyped/DefinitelyTyped#73770, I defined type declarations for `spf-check`.

In those declarations, I use `ipaddr.js` for definition of ip inputs. `ipaddr.js` is already a dependency of the package the types are for, and is also a dependency of my type declarations.

andrewbranch (approved):

PR #1206, opened 2025-10-04
Add `assertion-error` to allowed dependencies

MysteryBlokHed:
Related PR: DefinitelyTyped/DefinitelyTyped#73830

The `chai` package uses and re-exports the `AssertionError` class from this package.


jakebailey:
What version of chai does this? I checked latest and it doesn't list it as a dep, but maybe I missed something.

jakebailey (approved):

jakebailey:
Only in npm and older yarn, since they hoist. Anyway, thanks for explaining.

PR #1224, opened 2025-12-15
Support more than one project

christianliebel:
This PR adds support for more than one project in the `projects` array, concatenated into a single string.

### DIFF
--- a/packages/definitions-parser/test/packages.test.ts
+++ b/packages/definitions-parser/test/packages.test.ts
@@ -180,6 +180,42 @@ describe(TypingsData, () => {
     expect(data.isNotNeeded()).toBe(false);
   });
 
+  describe("project", () => {
+    it("returns the name of the only project", () => {
+      expect(data.projectName).toBe("zombo.com");
+    });
+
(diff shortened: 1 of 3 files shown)


packages/definitions-parser/test/packages.test.ts
```diff
@@ -180,6 +180,30 @@ describe(TypingsData, () => {
     expect(data.isNotNeeded()).toBe(false);
   });
 
+  describe("project", () => {
+    it("returns the name of the only project", () => {
+      expect(data.projectName).toBe("zombo.com");
+    });
+
+    it("returns the versioned name if not latest", () => {
```
tomayac: Should this return the version numbers? This line suggests it does, but the expected output is just the names of the projects.

tomayac (approved):
LGTM, thanks

jakebailey:
Since this touches a published package, it'll need a changeset. You can just do `pnpm changeset` and say it's a patch release.

That being said, this technically sticks display formatting into the parser. I think it would probably make more sense to leave `projectName`, add a new `projects` getter that returns the whole array, then use that in the code that produces the readme.

PR #1281, opened 2026-04-24
Generate tests for dtslint with CTA

JanKrivanek:


packages/dtslint/test/index.extra.test.ts
```diff
@@ -0,0 +1,30 @@
+/// <reference types="jest" />
+import { assertPackageIsNotDeprecated } from "../src/index";
+
+describe("assertPackageIsNotDeprecated extended", () => {
```
sandersn, line 4: why not edit the existing index.test.ts?

sandersn (commented):
This is basically the same as the vanilla one except with a lot less coverage. Still only covers pure functions, even though the high-value tests would be of the impure functions.

PR #1283, opened 2026-04-24
Generate tests in header-parser with Vanilla

JanKrivanek:


packages/header-parser/test/index.test.ts
```diff
@@ -174,3 +185,741 @@ Expected one of: ["MIT","Apache-2.0"]}`,
...
+    const result = validatePackageJson("mylib", makePkg({ nonNpm: true }), []);
+    expect(result).toEqual(
+      expect.arrayContaining([
+        expect.stringContaining('has missing "nonNpmDescription"'),
+      ]),
+    );
+  });
+
+  it("rejects nonNpm: true with non-string nonNpmDescription", () => {
+    const result = validatePackageJson("mylib", makePkg({ nonNpm: true, nonNpmDescription: 123 }), []);
```
sandersn, line 247: good coverage of edge cases

packages/header-parser/test/index.test.ts
```diff
@@ -174,3 +185,741 @@ Expected one of: ["MIT","Apache-2.0"]}`,
...
+      name: "@types/mylib",
+      version: "1.0.9999",
+      devDependencies: { "@types/mylib": "workspace:." },
+      owners: [{ name: "Test User", githubUsername: "testuser" }],
+      projects: ["https://example.com"],
+      ...overrides,
+    };
+  }
+
+  it("rejects missing projects", () => {
```
sandersn, line 324: this is one that's basic enough that I had trouble remembering what it was testing. So it's a good check-your-assumptions case.

packages/header-parser/test/index.test.ts
```diff
@@ -1,4 +1,15 @@
-import { validatePackageJson, makeTypesVersionsForPackageJson, License, getLicenseFromPackageJson } from "../src";
+import {
```
sandersn, line 1: don't reformat these; this repo is formatted by prettier (or dprint? can't remember)

sandersn (commented):
From memory, this looks like complete coverage of the properties we neglected to test. Those are low-value but it's nice to have  better coverage.

packages/header-parser/test/index.test.ts
sandersn: oh that's because there are more things imported now

PR #1299, opened 2026-06-09
Fix first time approvals, but disable, fix merge

jakebailey:
The app needs a new way to check for unapproved CI, otherwise it 403s on the daily PR sync.

But, it turns out that this simply never worked with the PAT, so I'm adding the code but turning it off and we can see if we want to reenable it later and auto approve even new account + new contributors.

The one PR that this affected was spam, so I think this feature should just die.


gabritto (dismissed):

gabritto (approved):

jakebailey:
Fixing merge perms while here

PR #1321, opened 2026-08-17
Add TypeScript 7 support to dtslint

jakebailey:
It turns out that we have enough API in 7.0 and 7.1 to be able to do `$ExpectType` and diagnostics. So, do that!


packages/dtslint/src/lintTypeScript7.ts
```diff
@@ -0,0 +1,443 @@
...
+      for (const configPath of configPaths) {
+        const project = snapshot.getProject(configPath);
+        if (!project) {
+          failures.push({
+            message: `TypeScript@${version} could not open ${configPath}.`,
+          });
+          continue;
+        }
+        failures.push(...getDiagnosticFailures(project, dirPath, version, isLatest));
+        failures.push(...getExpectTypeFailures(project, apiModule, astModule, dirPath, version, isLatest));
```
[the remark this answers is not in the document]
jakebailey: Fixed in ff76178c. TypeScript 7 failures are now aggregated by location/message across projects, with the producing version + tsconfig runs attached to both diagnostics and $ExpectType messages. Added a two-tsconfig fixture that verifies each identical failure is reported once and lists both configs.

packages/dtslint/src/lintTypeScript7.ts
```diff
@@ -0,0 +1,443 @@
...
+      const types = node.types
+        .map((item) => [item, print(item)] as const)
+        .sort((a, b) => a[1].localeCompare(b[1]))
+        .map(([item]) => item);
+      return ts.factory.updateUnionTypeNode(node, ts.factory.createNodeArray(types));
+    }
+    if (ts.isTypeLiteralNode(node)) {
+      const members = [...node.members].sort((a, b) => print(a).localeCompare(print(b)));
+      return ts.factory.updateTypeLiteralNode(node, ts.factory.createNodeArray(members));
+    }
```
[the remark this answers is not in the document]
jakebailey: Fixed in ff76178c. Type-literal members are now reordered only when every member is a property signature. Call signatures and overloaded methods preserve order; regression fixtures verify reversed overloads do not compare equal, while explicit $ExpectType alternatives using || remain supported.

packages/utils/src/expect-error-range.ts
```diff
@@ -0,0 +1,20 @@
+import * as semver from "semver";
+
+// Based on TypeScript's scanner.ts.
+const expectErrorSingleLine = /^\/\/\/?\s*@ts-expect-error\s+(.*)/;
+const expectErrorMultiLine = /^(?:\/|\*)*\s*@ts-expect-error\s+(.*?)\s*(?:\*\/)?$/;
+
+export function isVersionedExpectErrorOutsideRange(text: string, version: string): boolean {
+  const match = text.match(expectErrorSingleLine) || text.match(expectErrorMultiLine);
```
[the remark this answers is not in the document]
jakebailey: This is preexisting, and also, ironically you have not escaped this username, which is hilarious

jakebailey:
One tweak I need to make is to not call this "typescript 7"

packages/dtslint/src/lintCorsa.ts
```diff
@@ -0,0 +1,464 @@
...
+
+interface VariableStatement extends Node {
+  readonly declarationList: {
+    readonly declarations: readonly {
+      readonly initializer?: Node;
+    }[];
+  };
+}
+
+interface SourceFile extends Node {
```
andrewbranch: Can we not add TS 7 as a dev dependency for the shipped types?

jakebailey:
Going to hold off on this until TS 7.1 has `parseSourceFile`, then I will just use 6.0+7.0 together, with 7.1+ being Strada API free.

packages/dtslint/src/lintCorsa.ts
```diff
@@ -0,0 +1,522 @@
...
+import type { TsVersion } from "./lint";
+import { resolveLocalTypeScript } from "./typescript-installer";
+
+type CorsaApi = typeof import("typescript-7.1/unstable/sync");
+type CorsaAst = typeof import("typescript-7.1/unstable/ast");
+type CorsaFactory = typeof import("typescript-7.1/unstable/ast/factory");
+type CorsaApiInstance = InstanceType<CorsaApi["API"]>;
+interface LegacyCorsaApiInstance {
+  updateSnapshot(params: { openProjects: readonly string[] }): Snapshot;
+}
```
andrewbranch, line 17: Do we really need to support previous nightlies?

jakebailey: This is for 7.0 support, not specifically old nightlies. I feel wary not supporting 7.0 at all given 7.1 has type related changes in it.

andrewbranch: Hm, but _theoretically_ being 6.0 type-compatible is being 7.0 type-compatible, right? I feel like it takes a lot of workarounds to make this work with the 7.0 unstable API, but I guess it's not a ton of code.

jakebailey: Sadly, there are quite a few real differences that 7.0 already showed: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/75408

There's error positioning, which could be worked around, but there's also problems with es5 removal, strict binding, etc, which 6.0 still avoids.

I could imagine skipping 7.0 but then we'd have a hard time understanding how to typesVersions it.

PR #1327, opened 2026-09-01
Show package versions in mergebot comments

jakebailey:
Given the version isn't in index.d.ts anymore, I've frequently wanted to get to this info easily without clicking through files to find it.


gabritto (approved):
That's very useful, thanks!

PR #1328, opened 2026-09-01
Use test merge commits for package state

jakebailey:
If someone sends a PR from and old commit, then we'll get author data wrong. Use the GitHub generated merge commit instead.


jakebailey:
I think this is fine but it rarely happens that anyone sends a PR from an ancient fork, it just happened during my week and I thought it was fixable.

PR #1341, opened 2026-09-17
Add YAML pipeline for TypeScript DT runs

jakebailey:
Replacing our non-yaml one.


.changeset/major-boxes-sing.md
```diff
@@ -0,0 +1,5 @@
+---
```
jakebailey, line 1: The post-results script isn't a part of the package API so I'm ignoring the changeset

.azure-pipelines/scripts/run-typescript-pr-dt.sh
```diff
@@ -0,0 +1,132 @@
...
+
+install_typescript() {
+    if [[ -f package-lock.json ]]; then
+        npm ci
+    else
+        npm install
+    fi
+}
+
+install_go() {
```
weswigham: ...Why is installing go a shell script and not, like, step of the action environment setup? Just to handle PRs targeting an old `go` version that'd have issues when it merges with `main` anyway? Even then, pretty sure you could read the `go.mod` into an env variable in a prior step and pass it into the setup task, no?

jakebailey: These are actually all copy pasted from the current pipeline, which I reconfigured to work for Strada and Corsa. I'll just make this unconditionally set up go.

.azure-pipelines/scripts/run-typescript-pr-dt.sh
```diff
@@ -0,0 +1,114 @@
...
+        echo "$label SHA mismatch: expected $expected, got $actual" >&2
+        exit 1
+    fi
+}
+
+verify_sha merge "$actual_merge_sha" "$EXPECTED_MERGE_SHA"
+verify_sha base "$actual_base_sha" "$EXPECTED_BASE_SHA"
+verify_sha head "$actual_head_sha" "$EXPECTED_HEAD_SHA"
+
+install_typescript() {
```
weswigham: ...When would we not have a `package-lock.json` in the typescript repo...? A, like, 8 year old PR?

.azure-pipelines/scripts/run-typescript-pr-dt.sh
```diff
@@ -0,0 +1,114 @@
...
+    if [[ -f package-lock.json ]]; then
+        npm ci
+    else
+        npm install
+    fi
+}
+
+build_typescript() {
+    if [[ -d tsc ]]; then
```
weswigham: This is only needed to discern the old `typescript-go` layout from `typescript`, something I don't think we should need to do anymore in a task in `typescript`, right?

.azure-pipelines/scripts/run-typescript-pr-dt.sh
jakebailey: Yeah, this is all leftover from copy pasting; I'll just remove it all since the bot rejects PRs with merge conflicts and therefore can only work with corsa

.azure-pipelines/typescript-pr-dt.yml
```diff
@@ -0,0 +1,199 @@
...
+parameters:
+  - name: DT_SHA
+    type: string
+  - name: distinct_id
+    type: string
+  - name: expected_base_sha
+    type: string
+  - name: expected_head_sha
+    type: string
+  - name: expected_merge_sha
```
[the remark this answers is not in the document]
jakebailey: Correct, I can only do that after I make the pipeline and therefore have the pipeline ID.

weswigham (approved):

